PR #16, opened 2021-07-14
Fix spelling of search

RobertKrawitz:
Fix shearch -> search


ebattat (commented):
Thx, There are more place. I will fix it later

PR #211, opened 2021-08-18
Operator

ebattat:


benchmark_runner/common/oc/oc.py
```diff
@@ -318,10 +318,10 @@ def wait_for_initialized(self, label: str, workload: str = '', status: str = 'In
             if 'met' in result.decode("utf-8"):
                 return True
         except Exception as err:
-            if 'pod' in workload:
```
RobertKrawitz, line 321: At some point I think we want to not rely on the name of the workload to determine what it is, but this isn't worse than what it was prior.

RobertKrawitz:
Actually, could you collapse all of these commits into one so it's easier to understand?

PR #213, opened 2021-08-30
If kubeadmin password is not provided, assume that the cluster is already logged in

RobertKrawitz:
* If the kubeconfig is already logged in, it's not necessary to `oc login`, and that may fail on clusters provisioned on bare metal.
* Ensure that tests for timeout failures actually do time out, even if the underlying operation completes very quickly.
* Fix incorrect arguments to logging in ssh.py.

### DIFF
--- a/benchmark_runner/common/oc/oc.py
+++ b/benchmark_runner/common/oc/oc.py
@@ -167,12 +167,11 @@ class OC(SSH):
         """
 
         try:
-            if self.__kubeadmin_password:
-                return self.run(f'oc login -u kubeadmin -p {self.__kubeadmin_password}', is_check=True)
-            else:
-                raise LoginFailed
+            if self.__kubeadmin_password and self.__kubeadmin_password != '':
+                self.run(f'oc login -u kubeadmin -p {self.__kubeadmin_password}', is_check=True)
         except Exception as err:
             raise LoginFailed
+        return True
 
     @logger_time_stamp
     def get_pods(self):
--- a/benchmark_runner/common/ssh/ssh.py
+++ b/benchmark_runner/common/ssh/ssh.py
@@ -30,7 +30,7 @@ class SSH:
                 output = subprocess.getoutput(cmd)
             return output
         except subprocess.CalledProcessError as err:
-            logger.error("subprocess Status : FAIL", err.returncode, err.output)
+            logger.error(f'subprocess Status : FAIL: {err.returncode} {err.output}')
             raise SSHSubprocessError()
         except Exception as err:
              raise err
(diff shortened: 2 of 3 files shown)


ebattat (commented):
Need to delete LoginFailed totally in the following places:
1. Delete totally else section, no need return
2. Delete the error: LoginFailed
https://github.com/redhat-performance/benchmark-runner/blob/main/benchmark_runner/common/oc/oc.py#L6
3. Delete custom error:
https://github.com/redhat-performance/benchmark-runner/blob/main/benchmark_runner/common/oc/oc_exceptions.py#L8
4. Delete related test:
https://github.com/redhat-performance/benchmark-runner/blob/main/tests/unittest/benchmark_runner/common/oc/test_oc.py#L8
https://github.com/redhat-performance/benchmark-runner/blob/main/tests/unittest/benchmark_runner/common/oc/test_oc.py#L5
5. raise err instead of LoginFailed in https://github.com/redhat-performance/benchmark-runner/blob/main/benchmark_runner/common/oc/oc.py#L139

RobertKrawitz:
> Need to delete LoginFailed totally in the following places:
> 
> 1. Delete totally else section, no need return
> 2. Delete the error: LoginFailed
>    https://github.com/redhat-performance/benchmark-runner/blob/main/benchmark_runner/common/oc/oc.py#L6
> 3. Delete custom error:
>    https://github.com/redhat-performance/benchmark-runner/blob/main/benchmark_runner/common/oc/oc_exceptions.py#L8
> 4. Delete related test:
>    https://github.com/redhat-performance/benchmark-runner/blob/main/tests/unittest/benchmark_runner/common/oc/test_oc.py#L8
>    https://github.com/redhat-performance/benchmark-runner/blob/main/tests/unittest/benchmark_runner/common/oc/test_oc.py#L5
> 5. raise err instead of LoginFailed in https://github.com/redhat-performance/benchmark-runner/blob/main/benchmark_runner/common/oc/oc.py#L139

Shouldn't the login still fail if the wrong password is provided?

Perhaps there should be a test for oc login to verify that it fails if a bad password is provided.

RobertKrawitz:
I think there are a few ways we could handle this:

1) Keep the LoginFailed exception
2) Return True if login succeeds, False if it doesn't.

@ebattat what do you think?

ebattat:
Yes, I agree just return False. Just need just fix the unittest because its test empty pass.

RobertKrawitz:
Actually, on further thought I think that login failure /should/ be an exception -- it's normally expected to succeed, and if it doesn't, there's a problem.

RobertKrawitz:
Otherwise I have to modify all of the tests that call oc login to check the return value, which doesn't seem right.

RobertKrawitz:
Per discussion, will raise LoginFailed if the login fails, just as at present.

Added TestBadLogin test.

RobertKrawitz:
Also added fix for timeouts for operations that complete very quickly.

RobertKrawitz:
@ebattat squashed commits

ebattat:
/lgtm
/approve

Please squash and merge.

PR #239, opened 2021-09-23
add ci test against PR - ~15 min run

ebattat:
This PR contains:
1. Add PR Test CI [each PR push] - CI test against each PR that **run only 15 minutes**
2. Skip double tests
3. Split tests that didnt require benchmark-operator pod
4. Reduce the Test CI [merge to main] - run **~50 min instead of 60 min**


.github/workflows/PR_Test_CI.yml
```diff
@@ -0,0 +1,62 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: PR Test CI
+
+on:
+  pull_request:
+    branches: [ main ]
```
RobertKrawitz, line 8: What precisely does this mean?  Does it mean compare the PR against the main branch?  Because PRs won't be on the main branch.

.github/workflows/PR_Test_CI.yml
```diff
@@ -0,0 +1,62 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
RobertKrawitz, line 12: Why Ubuntu?  Why not CentOS or Fedora?

.github/workflows/PR_Test_CI.yml
```diff
@@ -0,0 +1,62 @@
...
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 pytest pytest-cov
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
RobertKrawitz, line 23: Best to use [[ ]] rather than [ ] with modern shells

.github/workflows/PR_Test_CI.yml
```diff
@@ -0,0 +1,62 @@
...
+        if [ -f tests_requirements.txt ]; then pip install -r tests_requirements.txt; fi
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - uses: azure/k8s-set-context@v1
```
RobertKrawitz, line 31: I thought we were running on IBM?

RobertKrawitz (changes requested):
Please fix the [ ] vs. [[ ]]; otherwise it's really questions.

.github/workflows/PR_Test_CI.yml
ebattat: That means is will run against each PR

.github/workflows/PR_Test_CI.yml
ebattat: Its the recommend light and agile machine for Linux

.github/workflows/PR_Test_CI.yml
ebattat: Fixed

.github/workflows/PR_Test_CI.yml
ebattat: Yes, it works for IBM so I use it

.github/workflows/PR_Test_CI.yml
RobertKrawitz: OK

.github/workflows/PR_Test_CI.yml
RobertKrawitz: Given that the real thing will be running on rhcos, I strongly urge you to consider using Fedora or CentOS as the base.  Any of these images will boot quickly enough.

.github/workflows/PR_Test_CI.yml
RobertKrawitz: What does this mean, then?

.github/workflows/PR_Test_CI.yml
ebattat: Those are the options: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idruns-on

.github/workflows/PR_Test_CI.yml
ebattat: Its take the Kubeconfig secret and pass it as environment variable
for more details read: https://github.com/Azure/k8s-set-context

.github/workflows/PR_Test_CI.yml
RobertKrawitz: OK, fair enough then.

RobertKrawitz (approved):

RobertKrawitz:
/approve
/lgtm
Please merge.

PR #249, opened 2021-10-15
Eliminate duplicate requirement

RobertKrawitz:
Eliminate duplicate requirement that causes pip3 tto error out


ebattat:
Hi,
This is the current file https://github.com/redhat-performance/benchmark-runner/blob/main/requirements.txt
What is the duplicate package ?

RobertKrawitz:
Done.

PR #253, opened 2021-10-19
Readthedoc

ebattat:
This Pr contains the following fixes:

1. stop_when_workload_finish: Add this variable for debug
2. SystemMetricsRequiredElasticSearch: raise this error if run system metrics without elasticsearch
3. OCSNonInstalled: raise this error when running 'OCS_PVC' and OCS operator is not installed
4. Add md support for readthedocs
5. Fix uperf label issue when missing PIN_NODE1


benchmark_runner/benchmark_operator/benchmark_operator_exceptions.py
```diff
@@ -5,3 +5,19 @@ class BenchmarkOperatorError(Exception):
     pass
 
 
+class OCSNonInstalled(BenchmarkOperatorError):
+    """
+    This class for run command error
+    """
+    def __init__(self):
```
RobertKrawitz, line 12: Can you just treat these as FALSE if this happens?

benchmark_runner/benchmark_operator/benchmark_operator_workloads.py
```diff
@@ -365,16 +391,24 @@ def uperf_pod(self):
             self.__oc.create_pod_sync(yaml=os.path.join(f'{self.__current_run_path}', f'{self.uperf_pod.__name__}.yaml'), pod_name=f'uperf-server')
             # uperf server
             server_name = self.__environment_variables_dict.get('pin_node1', '')
-            # name up to 27 chars
-            if len(server_name) > 27:
-                server_name = server_name[:28]
-            self.__oc.wait_for_initialized(label=f'app=uperf-bench-server-{server_name}-0', workload=workload)
-            self.__oc.wait_for_ready(label=f'app=uperf-bench-server-{server_name}-0', workload=workload)
+            if server_name:
+                # name up to 27 chars
```
RobertKrawitz, line 395: Why the fixed upper limit?

benchmark_runner/main/environment_variables.py
```diff
@@ -31,6 +31,23 @@ def __init__(self):
         self._environment_variables_dict['elasticsearch'] = os.environ.get('ELASTICSEARCH', '')
         self._environment_variables_dict['elasticsearch_port'] = os.environ.get('ELASTICSEARCH_PORT', '')
 
+        # This parameter 'True'/'False' for debug workload, when its set to 'True' it will stop the workload when finishing
+        self._environment_variables_dict['stop_when_workload_finish'] = os.environ.get('STOP_WHEN_WORKLOAD_FINISH', '')
+
+        # default parameter - change only if needed
+        # Parameters below related to 'run_workload()'
+        self._environment_variables_dict['workloads'] = ['stressng_pod', 'stressng_vm', 'uperf_pod', 'uperf_vm', 'hammerdb_pod_mariadb', 'hammerdb_vm_mariadb',  'hammerdb_pod_postgres', 'hammerdb_vm_postgres', 'hammerdb_pod_mssql', 'hammerdb_vm_mssql']
```
RobertKrawitz, line 39: This is really a constant.

benchmark_runner/main/environment_variables.py
```diff
@@ -31,6 +31,23 @@ def __init__(self):
...
+        self._environment_variables_dict['namespace'] = os.environ.get('NAMESPACE', 'benchmark-operator')
+        # run Hammerdb workload with ocs pvc True/False
+        self._environment_variables_dict['ocs_pvc'] = os.environ.get('OCS_PVC', 'True')
+        # This parameter get from Test_CI.yml file
+        self._environment_variables_dict['build_version'] = os.environ.get('BUILD_VERSION', '1.0.0')
```
RobertKrawitz, line 44: Isn't this also a constant?

benchmark_runner/benchmark_operator/benchmark_operator_workloads.py
ebattat: When pin_node is missing we got empty value. so I will skip it when missing pin node

benchmark_runner/main/environment_variables.py
RobertKrawitz: This isn't something the user can meaningfully modify, right?  They would need to add code to do so.

benchmark_runner/benchmark_operator/benchmark_operator_workloads.py
RobertKrawitz: That's not my question; why is there an uppoer limit of 27 characters?  The DNS limit is something like 250, as I recall.

benchmark_runner/main/environment_variables.py
```diff
@@ -31,6 +31,23 @@ def __init__(self):
...
 
+        # This parameter 'True'/'False' for debug workload, when its set to 'True' it will stop the workload when finishing
+        self._environment_variables_dict['stop_when_workload_finish'] = os.environ.get('STOP_WHEN_WORKLOAD_FINISH', '')
+
+        # default parameter - change only if needed
+        # Parameters below related to 'run_workload()'
+        self._environment_variables_dict['workloads'] = ['stressng_pod', 'stressng_vm', 'uperf_pod', 'uperf_vm', 'hammerdb_pod_mariadb', 'hammerdb_vm_mariadb',  'hammerdb_pod_postgres', 'hammerdb_vm_postgres', 'hammerdb_pod_mssql', 'hammerdb_vm_mssql']
+        self._environment_variables_dict['namespace'] = os.environ.get('NAMESPACE', 'benchmark-operator')
+        # run Hammerdb workload with ocs pvc True/False
+        self._environment_variables_dict['ocs_pvc'] = os.environ.get('OCS_PVC', 'True')
```
RobertKrawitz, line 42: Whereas this is a configuration option that it is meaningful for the user to set.

benchmark_runner/main/environment_variables.py
RobertKrawitz: This is also something the user should never have to set.

benchmark_runner/main/environment_variables.py
ebattat: There are defaults parameters for the CI runs but the user can change it for his needs.

benchmark_runner/benchmark_operator/benchmark_operator_workloads.py
ebattat: That is the limitation of Uperf workload in benchmark-wrapper, so I have set it accordingly.

RobertKrawitz: Then rather than just "name up to 27 chars", document it as "uperf workload is limited to 27 characters".  The "why" is more important than the "what".

ebattat: I dont have time to dig into it and understand why. I am focus on verifying that Uperf initialized properly

benchmark_runner/main/environment_variables.py
ebattat: User is not able to change this file at all. User needs just pass environment variable when running the container.
This file is for internal debugging

benchmark_runner/main/environment_variables.py
RobertKrawitz: All I'm asking is to structure the file so that options the user might set are clearly separate from those a developer might change, so the file better documents itself.

ebattat: I have added this comment:
# run Hammerdb workload with ocs pvc True/False. True=OCS, False=Ephemeral

PR #254, opened 2021-10-20
update performance metrics for hammerdb

ebattat:
I have updated hammerdb performance metric for nightly build according to documentation


ebattat:
@jeniferh, Can you please review the performance metrics before merge it ?

benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/hammerdb_data_template.yaml
```diff
@@ -1,7 +1,7 @@
 shared_data:
   namespace: {{ namespace }}
-  db_num_workers: 2
-  db_warehouses: 2
+  db_num_workers: 24
```
jeniferh, line 3: @ebattat the scale doc has the following number of users aka db_num_workers (each a diff test):
Num users: 1, 2, 4, 8,16, 24

ebattat: Yes, It will run all this users: 1, 2, 4, 8,16, 24

jeniferh: > Yes, It will run all this users: 1, 2, 4, 8,16, 24
Where do the lower user counts get set? 
> Do you know where we can configure it
Double check w/ Sanjay but I believe that was a param that needs to be passed as part of the db config values (i.e. configured in each db conf file)

ebattat: @jeniferh OK, I will merge it to have a nightly run with current performance configuration.
Once I will get Sanjay response I will fix it on other PR if needed

PR #264, opened 2021-10-25
Change VM vcpu topology from cores to sockets 

jeniferh:
"Sockets" are the default vcpu topology in KVM for perf reasons (uses a scheduler sync_wake hint that helps certain types of workloads)


ebattat (commented):
Need to fix it to strssng and hammerdb for all flavors. Can you pls fix it?

jeniferh:
I am going to close this for now and make a broader sockets topo change once PR#278 lands

PR #269, opened 2021-10-29
Various updates to developer doc

RobertKrawitz:
Based on development experience with kata


ebattat (commented):
Do you want to add also to run requirement file

PR #271, opened 2021-11-01
separate db CPU and memory and change mssql CPU to 32

ebattat:
Change MSSQL CPU to 32

### DIFF
--- a/benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/hammerdb_data_template.yaml
+++ b/benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/hammerdb_data_template.yaml
@@ -18,6 +18,8 @@ shared_data:
   metrics_profile: node-metrics.yml
   #metrics_profile: https://raw.githubusercontent.com/cloud-bulldozer/benchmark-operator/master/roles/kube-burner/files/metrics.yaml
 mariadb:
+  mariadb_limits_cpu: 16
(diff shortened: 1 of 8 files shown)


benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/hammerdb_data_template.yaml
```diff
@@ -18,20 +18,26 @@ shared_data:
...
+  mssql_limits_memory_GB: 16
   db_image: centos8-mssql2019-container-disk:latest
   db_pass: s3curePasswordString
   db_port: 1433
   db_type: mssql
   db_user: SA
   db_server_pod: mssql-deployment.mssql-db
 postgres:
+  postgres_limits_cpu: 16
+  postgres_limits_memory_GB: 16
```
RobertKrawitz, line 40: Am I understanding correctly that there was no limit prior?  That would result in an implicit limit for Kata of 2 GB RAM and give a single core to the VM.

RobertKrawitz (approved):

benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/hammerdb_data_template.yaml
ebattat: There was a prior limitation it was 
limits_cpu: 16
memory_GB: 16

PR #272, opened 2021-11-02
update hammerdb perf metrics

ebattat:
I have update all hammerdb conf to use the following:
CPU: 32
Memory: 64GB

### DIFF
--- a/benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/internal_data/hammerdb_vm_template.yaml
+++ b/benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/internal_data/hammerdb_vm_template.yaml
@@ -74,16 +74,16 @@ spec:
       kind: vm
       client_vm:
         dedicatedcpuplacement: false
-        sockets: 1
-        cores: {{ cores }}
+        sockets: {{ vm_sockets }}
+        cores: {{ vm_cores }}
         threads: 1
         image: quay.io/cloud-bulldozer/{{ db_image }}
         limits:
-          memory: {{ limits_memory_GB }}Gi
+          memory: {{ vm_limits_memory_GB }}Gi
         requests:
-          memory: 10Mi
+          memory: {{ vm_requests_memory_GB }}Gi
         network:
-          front_end: bridge # or masquerade
+          front_end: masquerade
           multiqueue:
             enabled: false # if set to true, highly recommend to set selinux to permissive on the nodes where the vms would be scheduled
             queues: 0 # must be given if enabled is set to true and ideally should be set to vcpus ideally so sockets*threads*cores, your image must've ethtool installed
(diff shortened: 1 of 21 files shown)


RobertKrawitz (approved):

benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/internal_data/hammerdb_vm_template.yaml
```diff
@@ -74,14 +74,14 @@ spec:
...
         dedicatedcpuplacement: false
-        sockets: 1
-        cores: {{ cores }}
+        sockets: {{ vm_sockets }}
+        cores: {{ vm_cores }}
         threads: 1
         image: quay.io/cloud-bulldozer/{{ db_image }}
         limits:
-          memory: {{ limits_memory_GB }}Gi
+          memory: {{ vm_limits_memory_GB }}Gi
```
jeniferh, line 82: I would recommend removing memory limits for VMs as it can cause OOM if qemu mem overhead is a bit too high. In general I don't think we need to set limits for VMs specifically because they are already bound by the guest size (unlike pods/kata).

benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/internal_data/hammerdb_vm_template.yaml
```diff
@@ -74,14 +74,14 @@ spec:
...
         requests:
-          memory: 10Mi
+          memory: {{ vm_requests_memory_GB }}Gi
```
jeniferh, line 84: May want to add cpu requests for the VM as well otherwise you are only guaranteed 1/10th of a cpu worth of shares. If you are making sockets and cores both configurable, the total cpu requests would be sockets * cores.

benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/internal_data/hammerdb_vm_template.yaml
```diff
@@ -74,14 +74,14 @@ spec:
...
         network:
           front_end: bridge # or masquerade
```
jeniferh, line 86: Could you change to masquerade? It is the default and bridge is unsupported.

jeniferh (commented):
I made some comments on the VM req/limits but otherwise looks good!

benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/internal_data/hammerdb_vm_template.yaml
ebattat: OK. good point I will remove it and test.

benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/internal_data/hammerdb_vm_template.yaml
ebattat: Yes, I configure the cpu and socket accordingly

benchmark_runner/benchmark_operator/workload_flavors/perf_ci/hammerdb/internal_data/hammerdb_vm_template.yaml
ebattat: Yes I will

PR #278, opened 2021-11-04
First round of template simplification: separate generation of data files from actual creation

RobertKrawitz:
Generate the benchmark and pod YAML files separately from writing them out.  The goal here is to be able to generate golden files without having to extract them from the current_run directories.  This also avoids generation of some temporary files that have to be cleaned up when they are only used for generating intermediate YAML that's simply a dictionary for creating payload YAML files.

Still want to clean up all uses of update_environment_variable and delete_generate_file to use the new APIs in update_data_template_yaml_with_environment_variables.py and remove those functions.


RobertKrawitz:
Golden file comparison passed with latest rebase.

benchmark_runner/benchmark_operator/workload_flavors/generate_yaml_from_workload_flavors.py
```diff
@@ -40,112 +61,94 @@ def __get_yaml_template_by_workload(self, workload: str, extension='.yaml', skip
...
         """
-        # replace environment variables and generate hammerdb_data.yaml
-        if 'kata' in workload:
-            environment_variables.environment_variables_dict['kind'] = 'kata'
+        if '_pod' in workload:
+            return 'pod', 'pod'
+        elif '_kata' in workload:
+            return 'kata', 'pod'
+        elif '_vm' in workload:
+            return 'vm', 'vm'
```
ebattat, line 78: Why we need to return couples ?

ebattat (commented):
Can you please rebase it and rerun pytest against Func env ?

PR #279, opened 2021-11-04
update hammerdb mem to 32GB

ebattat:
Decrease hammerdb database memory to 32GB for 64GB due to failure in kata


benchmark_runner/main/environment_variables.py
```diff
@@ -112,7 +112,7 @@ def __init__(self):
         # remote ssh timeout - 3 hours for installation time
         self._environment_variables_dict['provision_timeout'] = os.environ.get(f'{self.__ocp_env_flavor}_PROVISION_TIMEOUT', '10800')
         # General timeout - 1.5 hours wait for pod/vm/upload data to elasticsearch
-        self._environment_variables_dict['timeout'] = os.environ.get(f'{self.__ocp_env_flavor}_TIMEOUT', '5400')
+        self._environment_variables_dict['timeout'] = os.environ.get(f'{self.__ocp_env_flavor}_TIMEOUT', '3600')
```
RobertKrawitz, line 115: Still concerned about this, but if we hit it, we can re-evaluate what's going on.

RobertKrawitz (approved):

benchmark_runner/main/environment_variables.py
ebattat: We need to decide if we want to wait 1 hour for pod. I think that its too much

PR #293, opened 2021-11-18
Fix error in Kata workaround

RobertKrawitz:
Resources that end in .sh should not be treated as YAML.

### DIFF
--- a/benchmark_runner/common/oc/oc.py
+++ b/benchmark_runner/common/oc/oc.py
@@ -664,23 +664,24 @@ class OC(SSH):
         """
         for resource in resource_list:
             logger.info(f'run {resource}')
-            self._create_async(yaml=os.path.join(path, resource))
-            # for first script wait for virt-operator
-            if '01_operator.yaml' == resource:
-                # Wait that cnv operator will be created
-                self.wait_for_ocp_resource_create(resource='kata',
-                                                  verify_cmd='oc -n openshift-sandboxed-containers-operator wait deployment/controller-manager --for=condition=Available',
-                                                  status='deployment.apps/controller-manager condition met')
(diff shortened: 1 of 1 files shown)


benchmark_runner/common/oc/oc.py
```diff
@@ -664,23 +664,24 @@ def create_kata(self, path: str, resource_list: list):
...
-                                                  status='false')
-                total_nodes_count = self.run(cmd="oc get kataconfig -ojsonpath='{.items[0].status.total_nodes_count}'")
-                completed_nodes_count = self.run(cmd="oc get kataconfig -ojsonpath='{.items[0].status.installationStatus.completed.completed_nodes_count}'")
-                if total_nodes_count != completed_nodes_count:
-                    raise KataInstallationFailed(f'not all nodes installed successfully total {total_nodes_count} != completed {completed_nodes_count}')
-            #
-            elif '03_ocp48_patch.sh' == resource:
-                self.run(cmd=f'chmod +x {os.path.join(path, resource)}; {path}/./{resource}')
+            if resource.endswith('.sh'):
+                if '03_ocp48_patch.sh' == resource:
```
ebattat, line 668: Can you change the order to 1 or it must be 3 ?

RobertKrawitz: Must be 3.  This one has to be run last, only after everything else is done.  I could (and probably should) change the test so that only .yaml files would get fed to oc, which would have the same effect.

ebattat: Can we shift it to the end ?

RobertKrawitz:
Manual testing of this (via create_kata()) succeeded on a standalone bare metal cluster.  The nodes did get updated correctly.

PR #298, opened 2021-11-20
update ocp install

ebattat:
This PR split ipi installer to 2 steps:
1. run ocp installer
2. verify ocp installer

The reason for it is that need to run a separate container after running remote ocp installer ansible


ebattat (commented):
/LGTM

PR #301, opened 2021-12-01
Fix string to get kata namespace

RobertKrawitz:
As the string to retrieve the namespace contains a backslash, it should be a raw string.


ebattat (commented):
/LGTM
/Approved

PR #302, opened 2021-12-01
Remove stray YAML nits to make cleaner YAML files for the CI flavor collapse work

RobertKrawitz:
A lot of the YAML templates had unnecessary `---` lines at the top or bottom that make the CI flavor collapse work more difficult because they generate spurious changes.  This removes them.  Currently in testing on func.


ebattat (commented):
/LGTM
/Approved

PR #304, opened 2021-12-08
Merge hammerdb and other workload template generation

RobertKrawitz:
Simplify the YAML generation by merging the hammerdb and other YAML generators into one routine.  Intent is that this will help simplify the process of writing other workloads in the future.

Passes golden file testing.  Currently testing on func_ci environment.


ebattat (commented):
/LGTM
/Approved

PR #309, opened 2021-12-13
Run artifacts

ebattat:
Run artifacts:

For every workload:

1. Capture workload pod/vm logs
2. Capture system-metrics pod log
3. Capture benchmark-operator pod log
4. Capture benchmark-runner log
5. tar.gz all the logs
6. Upload to IBM Cloud Object Storage (S3) with timestamp
7. Pytest for IBM Cloud Object Storage (S3) 

* Support only on functional environment

** Need to add the following:
1. Prometheus snapshot
2. Perf environment support


benchmark_runner/benchmark_operator/benchmark_operator_workloads.py
```diff
@@ -228,26 +231,28 @@ def tear_down_vm_after_error(self, yaml: str, vm_name: str):
...
         """
         This method run system metrics collector
-        @param workload:
+        @param workload: the workload
         :return:
         """
+        if self.__run_type == 'test_ci':
+            es_index = 'system-metrics-test-ci'
+        else:
+            es_index = 'system-metrics'
```
RobertKrawitz, line 240: Why is this pulled out of the `if self.es_host`?  This isn't needed if we're not uploading data.

benchmark_runner/benchmark_operator/benchmark_operator_workloads.py
```diff
@@ -295,43 +304,145 @@ def update_ci_status(self, status: str, ci_minutes_time: int, benchmark_operator
...
+        path = self.__environment_variables_dict.get('run_artifacts_path', '')
+        # verify that its not empty path
+        if len(path) > 3 and path != '/' and path and tar_run_artifacts_path and os.path.isfile(tar_run_artifacts_path):
+            # remove run_artifacts_path
+            shutil.rmtree(path=self.__environment_variables_dict.get('run_artifacts_path', ''))
+            # remove tar.gz file
+            os.remove(path=tar_run_artifacts_path)
+
+
+
```
RobertKrawitz, line 405: This is at lesat mostly common code, not benchmark-operator code.  Don't refactor it right now, but this will be one more thing we need to refactor.

benchmark_runner/benchmark_operator/benchmark_operator_workloads.py
```diff
@@ -295,43 +304,145 @@ def update_ci_status(self, status: str, ci_minutes_time: int, benchmark_operator
...
 #***********************************************************************************************
 ######################################## Workloads #############################################
 #***********************************************************************************************
 
+
     @typechecked
     @logger_time_stamp
-    def stressng_pod(self, name: str=''):
+    def stressng_pod(self, name: str = ''):
```
RobertKrawitz, line 413: Nit: I believe that there is not supposed to be whitespace in this usage.

benchmark_runner/benchmark_operator/benchmark_operator_workloads.py
```diff
@@ -295,43 +304,145 @@ def update_ci_status(self, status: str, ci_minutes_time: int, benchmark_operator
...
             if name == '':
                 name = self.stressng_pod.__name__
             workload = name.replace('_', '-')
             kind = 'pod'
             if '_kata' in name:
                 kind = 'kata'
+            if self.__run_type == 'test_ci':
+                es_index = 'stressng-test-ci-results'
+            else:
+                es_index = 'stressng-results'
```
RobertKrawitz, line 428: Again, why is this moved outside of the `if self.__es_host` block?

benchmark_runner/benchmark_operator/benchmark_operator_workloads.py
```diff
@@ -422,30 +548,32 @@ def uperf_pod(self, name: str=''):
...
             self.__oc.wait_for_initialized(label='app=uperf-bench-client', workload=workload)
             self.__oc.wait_for_ready(label='app=uperf-bench-client', workload=workload)
-            self.__oc.wait_for_pod_completed(label='app=uperf-bench-client', workload=workload)
+            status = self.__oc.wait_for_pod_completed(label='app=uperf-bench-client', workload=workload)
+            status = 'complete' if status else 'failed'
             # system metrics
             if environment_variables.environment_variables_dict['system_metrics'] == 'True':
                 self.system_metrics_collector(workload=workload)
+            # save run artifacts logs
+            run_artifacts_url = self.__create_run_artifacts(workload=workload, labels=['uperf-client', 'uperf-server'])
```
RobertKrawitz, line 560: I'd really like this to be hoisted from the individual workload functions into `run_workload_func`.  It will make adding the metrics much easier and save a lot of code.

benchmark_runner/common/clouds/shared/s3/s3_operations.py
```diff
@@ -0,0 +1,233 @@
...
+        :return:
+        """
+        try:
+            self.__s3_client.upload_file(Filename=file_name_path,
+                                         Bucket=bucket,
+                                         Key=f'{key}/{upload_file}',
+                                         ExtraArgs={'ServerSideEncryption': 'AES256'})
+
+        except ClientError:
+            raise
```
RobertKrawitz, line 52: Missing exception?

benchmark_runner/main/environment_variables.py
```diff
@@ -50,6 +50,9 @@ def __init__(self):
         self._environment_variables_dict['run_types'] = ['test_ci', 'func_ci', 'perf_ci']
         # Run type test_ci/func_ci/perf_ci, default test_ci same environment as func_ci
         self._environment_variables_dict['run_type'] = os.environ.get('RUN_TYPE', 'test_ci')
+        # Benchmark runner local run artifacts path with time stamp format
+        self._environment_variables_dict['time_stamp_format'] = os.path.join(datetime.datetime.fromtimestamp(time.time()).strftime('%Y-%m-%d-%H-%M-%S'))
```
RobertKrawitz, line 54: I'd really like the timestamp to contain the timezone.  That's going to be needed to make sense of the prometheus snapshot.

RobertKrawitz (changes requested):
Two main issues:

1. I'd like the snapshot to happen higher up in the stack, because for the prom db I need to take the snapshot only after all the pods have been taken down.
2. I'd like the timestamp format to contain a timezone (%z), because to make sense of the prometheus snapshot it's important to have the timezone correct.

benchmark_runner/benchmark_operator/benchmark_operator_workloads.py
RobertKrawitz: Upon discussion, this can be done by wrapping `run_workload_func` so there is no issue here.

RobertKrawitz (approved):
My important issues are addressed; the design and code are fine as is.

benchmark_runner/benchmark_operator/benchmark_operator_workloads.py
ebattat: Fix Python PEP-8 comment

benchmark_runner/benchmark_operator/benchmark_operator_workloads.py
ebattat: Solved empty index issue in case of Exception.
In case of failure while running pod/vm, we didnt reach to self.__es_host and move direct to Exception for upload failure date into ES. in that case we got empty index. That the reason I moved it to the beginning.

benchmark_runner/common/clouds/shared/s3/s3_operations.py
ebattat: We dont want to catch it, we just move it forward to get more error details

PR #311, opened 2021-12-14
add run artifact fixes and setup update

ebattat:


benchmark_runner/main/environment_variables.py
```diff
@@ -53,6 +53,7 @@ def __init__(self):
         # Run type test_ci/func_ci/perf_ci, default test_ci same environment as func_ci
         self._environment_variables_dict['run_type'] = os.environ.get('RUN_TYPE', 'test_ci')
         # Benchmark runner local run artifacts path with time stamp format
+        self._environment_variables_dict['date_key'] = datetime.datetime.now().strftime("%Y/%m/%d")
```
RobertKrawitz, line 56: Needs a comment; it looks earlier like this is being used as part of a filename.

RobertKrawitz (commented):
I don't see the nature of the bug you're fixing with this.

benchmark_runner/main/environment_variables.py
ebattat: Updated

RobertKrawitz: "Path" suggests a Linux path, where the `/` will be path separators.  If that's actually the case, os.path.join() should be used on the individual components.  If this is simply an identifier that's not part of a POSIX path, that should be noted.

ebattat: I will add 'not part of a POSIX path'

RobertKrawitz: OK

RobertKrawitz (approved):

RobertKrawitz (approved):

PR #313, opened 2021-12-14
add init to shared folder and add missing exceptions

ebattat:
Add init to shared path and add missing s3 custom exception


RobertKrawitz (approved):
approve

PR #316, opened 2021-12-14
Net_queues for uperf VMs

jeniferh:
Enable network multiqueue for VMs:
 - Set num queues to total vcpus (this is the default for CNV VM templates)
 - Removed the note about SElinux since I did not see any issues with nodes set to 'Enforcing'
Also a small fix to switch one last 'bridge' net type (no longer supported) to masquerade.


jeniferh:
Thanks for the reviews. Just confirmed w/ a manual Perf cluster run the perf_ci config works properly and both VMs get 8 queues as expected.

PR #318, opened 2021-12-14
Stressng perf_ci "large" config

jeniferh:
- Change VM vcpu topology to sockets
 - Add perf_ci run type to stress "full" system (~75% host cpu, ~50% host mem, 5min run)

### DIFF
--- a/benchmark_runner/benchmark_operator/workload_flavors/templates/stressng/internal_data/stressng_vm_template.yaml
+++ b/benchmark_runner/benchmark_operator/workload_flavors/templates/stressng/internal_data/stressng_vm_template.yaml
@@ -36,8 +36,8 @@ spec:
       kind: vm
       client_vm:
         dedicatedcpuplacement: false
-        sockets: 1
-        cores: {{ cores }}
+        sockets: {{ sockets }}
+        cores: 1
         threads: 1
         image: quay.io/kubevirt/fedora-container-disk-images:34
         limits:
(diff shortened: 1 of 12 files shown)


jeniferh:
I tested on the Perf cluster, pod and vm types run fine for the new large perf_ci config but kata had a problem, it scheduled fine but appeared to die before the run completed, I will need to dig into it more tomorrow..

RobertKrawitz (approved):
Looks fine on the form; I'm going to trust you on the content.

ebattat (commented):
/LGTM
Need to rebase before merging

ebattat (commented):
/LGTM
/Approved

ebattat (approved):
/LGTM
/Approved

jeniferh:
Thanks. Kata was dying prematurely for larger stressng vm_bytes values (I'd like to eventually understand why..) but I did some comparisons and larger values don't impact the mem bogomips result much, so just dropping it to stress ~50% of host memory which passes fine.

RobertKrawitz:
> Thanks. Kata was dying prematurely for larger stressng vm_bytes values (I'd like to eventually understand why..) but I did some comparisons and larger values don't impact the mem bogomips result much, so just dropping it to stress ~50% of host memory which passes fine.

What exactly does vm_bytes do?  It might have been getting oom killed (by default, kata VMs have 2GiB).

jeniferh:
> What exactly does vm_bytes do? It might have been getting oom killed (by default, kata VMs have 2GiB).
It is the amount the test mmaps, but it should have had plenty of room since the pod limit was significantly larger.. so a bit puzzling.

RobertKrawitz:
How much?

jeniferh:
The failure was vm_bytes=75G w/ 90Gi req/limit. What is working now is vm_bytes=60G w/ 75Gi req/limit. In the failure case, the kata pod would get scheduled, start the test & mem allocation but it would just go into pod 'Complete' state much earlier than the 5min runtime and the pod logs had no additional info. I didn't have a chance yet to catch exactly what amount of allocation triggered it, and probably need to figure out where the kata qemu logs reside as well.

PR #321, opened 2021-12-15
disable s3 bucket default save

ebattat:
disable s3 bucket and also add official version of kubevirt with capturing fix

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,18 +30,10 @@ RUN  curl -L https://github.com/openshift/okd/releases/download/${oc_version}/op
      && cp ~/oc /usr/local/bin/oc
 
 # install virtctl for VNC
-# ARG virtctl_version=0.34.2
-# RUN curl -L https://github.com/kubevirt/kubevirt/releases/download/v${virtctl_version}/virtctl-v${virtctl_version}-linux-amd64 -o  ~/virtctl \
-#     && chmod +x ~/virtctl \
-#     && cp ~/virtctl /usr/local/bin/virtctl
-
-# Patch for capture vm: install GO & clone kubevirt & build virtctl
-RUN dnf install -y epel-release
-RUN dnf install -y go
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -30,18 +30,10 @@ RUN  curl -L https://github.com/openshift/okd/releases/download/${oc_version}/op
...
-RUN dnf install -y epel-release
-RUN dnf install -y go
-RUN git clone https://github.com/kubevirt/kubevirt ~/kubevirt \
-    && cd ~/kubevirt/cmd/virtctl \
-    && go build . \
-    && cp  ~/kubevirt/cmd/virtctl/virtctl /usr/local/bin/virtctl
+ARG virtctl_version=0.48.1
+RUN curl -L https://github.com/kubevirt/kubevirt/releases/download/v${virtctl_version}/virtctl-v${virtctl_version}-linux-amd64 -o  ~/virtctl \
+    && chmod +x ~/virtctl \
+    && cp ~/virtctl /usr/local/bin/virtctl
```
RobertKrawitz, line 36: @jeniferh could you take a look at this part?

RobertKrawitz (approved):
I'm OK with everything except the Dockerfile; @jeniferh should take a look at that.

Dockerfile
jeniferh: Yep, looks good, v0.48.1 will pick up the terminal setting fix (https://github.com/kubevirt/kubevirt/pull/6902)

ebattat: @jeniferh Right, Still I have a problem. Capture a vm log inside a benchmark-runner did not work. I guess that the root cause is redirect to dev null [here](https://github.com/redhat-performance/benchmark-runner/blob/c0ed3f4909fc897939b16b9a28e1abfac24ed3d7/benchmark_runner/common/ssh/ssh.py#L28)
I will delete it and hope it will solve the issue.

ebattat:
Add it to PR: vm_capture_fix

PR #323, opened 2021-12-16
add e2e test to PR push by running main

ebattat:
Add E2E test to every push PR running stressng pod workload


RobertKrawitz (approved):
lgtm

PR #326, opened 2021-12-17
fix capture vm

ebattat:
Fix for capture vm:
1. Mount local disk to running container 
2. Fix small issue with Prometheus
3. Fix E2E test by adding -v local disk


RobertKrawitz (approved):
lgtm

PR #355, opened 2022-01-03
fix major issue in system metrics uperf

ebattat:
Major issue for system metrics in uperf


RobertKrawitz (approved):
approve

ebattat:
Uperf workload is using uperf_ts instead of timestamp in ElasticSearch.
The problem is that system metrics using timestamp and not uperf_ts after uperf workload running.

PR #365, opened 2022-01-06
- Fix use of OCS_PVC=false

RobertKrawitz:
- Enable OCS_PVC=false for vdbench
- Enable no pin node for vdbench
(I just rebased this)


ebattat (approved):
/LGTM

PR #372, opened 2022-01-18
Update golden files, fix doc, and restore golden files check

RobertKrawitz:
The golden files check should not have been disabled.  Re-enabling.


RobertKrawitz:
Fixed by #373

PR #377, opened 2022-01-21
Fix upload es

ebattat:
This PR should fix vdbench upload data to elasticsearch issue.
Solved it by splitting nested dictionary to list of dictionaries


generate_golden_files.sh
```diff
@@ -1,3 +1,3 @@
-python -m venv venv
+python3 -m venv venv
```
RobertKrawitz, line 1: Is this only within the context of the script, or does it change any activated venv outside of the script?

RobertKrawitz: confirmed yes.

RobertKrawitz (approved):

PR #380, opened 2022-02-08
Wait for CRD to exist rather than named deployment for kata install

RobertKrawitz:
Rather than waiting for a deployment name that may change across releases, wait for the kataconfigs CRD to exist before attempting to create the kataconfig.


ebattat (approved):
/LGTM
==> Once we have func env I will merge it

RobertKrawitz:
Updated, please re-review.

ebattat (approved):
LGTM

PR #386, opened 2022-03-03
Work around CPUs offline issue for Kata

RobertKrawitz:
Kata in 4.9 and 4.10 has a problem with hotplugging CPUs whereby CPUs added by means of a CPU resource limit don't all get onlined.  This workaround is to use a pod annotation to change the default_vcpus setting to boot the Kata VM with the desired CPU count.  Most software works, possibly with a loss of performance, but mssql does not start.

xref https://github.com/kata-containers/kata-containers/issues/3793
xref https://github.com/microsoft/mssql-docker/issues/162

Caveats (at present):

1) Annotations in benchmark-operator workloads do not appear to be functioning correctly at present; I'm attempting to determine whether that's due to my error or a problem in benchmark-operator.

2) CI will need an update to set KATA_CPUOFFLINE_WORKAROUND=True


benchmark_runner/main/environment_variables.py
```diff
@@ -37,6 +37,10 @@ def __init__(self):
         self._environment_variables_dict['elasticsearch_user'] = os.environ.get('ELASTICSEARCH_USER', '')
         self._environment_variables_dict['elasticsearch_password'] = os.environ.get('ELASTICSEARCH_PASSWORD', '')
 
+        # Workaround for Kata CPU offline problem in 4.9/4.10
+        # Set to True to
+        self._environment_variables_dict['kata_cpuoffline_workaround'] = os.environ.get('KATA_CPUOFFLINE_WORKAROUND', '')
```
ebattat, line 42: Do you mean we need to enable it for all our workloads run ?

RobertKrawitz: We should enable it in the CI runs, and then when we have a proper fix, take it out.

ebattat: For all the workloads, right ?

ebattat:
Need to enable the environment variables in the following Github actions workflows:
-e KATA_CPUOFFLINE_WORKAROUND='True'

`https://github.com/redhat-performance/benchmark-runner/blob/afad66aef21055916b329e3e24f2c0692f2c3729/.github/workflows/Nightly_Func_Env_CI.yml#L108`
Add it after:
-e RUN_TYPE='func_ci'

`https://github.com/redhat-performance/benchmark-runner/blob/afad66aef21055916b329e3e24f2c0692f2c3729/.github/workflows/Nightly_Func_Env_Operator_CI.yml#L282`
Add it after:
-e RUN_TYPE='func_ci'

`https://github.com/redhat-performance/benchmark-runner/blob/afad66aef21055916b329e3e24f2c0692f2c3729/.github/workflows/Nightly_Perf_Env_CI.yml#L140`
Add it after:
-e RUN_TYPE='perf_ci'

`https://github.com/redhat-performance/benchmark-runner/blob/afad66aef21055916b329e3e24f2c0692f2c3729/.github/workflows/Nightly_Perf_Env_Operator_CI.yml#L282`
Add it after:
-e RUN_TYPE='perf_ci'

Also need to rebase branch

RobertKrawitz:
Yep, starting work on it.

RobertKrawitz:
Rebased and updated.

ebattat (approved):
/LGTM

PR #387, opened 2022-03-10
Stressng sequential change

jeniferh:
Changing run type to sequential instead of parallel, tests have proven results are much more stable using this mode.
Also reduced the runtime after confirming memory allocation and run variance looks fine for a 2min run for each test type.

### DIFF
--- a/benchmark_runner/common/template_operations/templates/stressng/stressng_data_template.yaml
+++ b/benchmark_runner/common/template_operations/templates/stressng/stressng_data_template.yaml
@@ -3,7 +3,7 @@ metadata:
 template_data:
   shared:
     pin_node: {{ pin_node1 }}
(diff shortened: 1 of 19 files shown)


jeniferh:
Note: tested and passed on PerfCI cluster

RobertKrawitz (approved):
lgtm.

It might be useful to have a parallel run separately, to see what happens with the variation over time.  Perhaps we could use various tools to analyze what's going on internally.  But that's a separate issue.

jeniferh:
Yeah I do think parallel is interesting, maybe something to add later, but the run to run variance swings in the memory tests are so wild that we need to figure out what is useful from a regression standpoint..

ebattat (approved):
/LGTM

PR #382, opened 2022-02-24
CNV nightly and ODF support

ebattat:
This PR include the following:

1. Add support for auto install CNV nightly build 
2. Add support for auto install ODF stable build [OCS => ODF]
3. Readthedocs update the method
4. Add custom install resource support - for running the COREDNS patch
5. Add Func env IPI + Operator auto weekly install
6. Add matrix strategy support for Perf and Func Nightly CI
7. Create a dedicate class 'create_ocp_resource_operations.py' for create ocp resource

PR #390, opened 2022-03-14
Fix Kata installation

RobertKrawitz:
PR #381 was improperly removed when PR #382 was merged.


ebattat (commented):
Need to bump tag version by 1 in setup.py and .bumpversion.cfg

ebattat (approved):
/LGTM

PR #391, opened 2022-03-15
Various enhancements

kpouget:
Sharing the enhancements I had to do on `benchmark-runner` to get my workload running. To be discussed.

-* `template_operations: do not crash if 'shared' is missing in the template`

---

* `environment_variables: read settings from an environment file`

---

* `workloads_operations: make 'oc' a class singleton`

---

* `benchmark_runner/common/ssh/ssh_exceptions.py: fix typo`

---

* `benchmark_runner/common/oc/oc.py: _create_async: raise an excption if the command fails`

---

* `benchmark_runner/main/environment_variables.py: allow overriding the namespace with env var`

---

* `benchmark_runner/main: allow setting the RUNNER_TYPE with an env variable`

---

* `benchmark_runner: allow reading the config from the CLI args`

---

* `benchmark_runner/common/template_operations/template_operations.py: allow storing the template data in the workload directory`

---

* `benchmark_runner: allow overriding initialize/finalize_workload methods`

---

* `benchmark_runner/main/environment_variables.py: allow setting RUN_ARTIFACTS_PATH`

---

* `benchmark_runner: report benchmark failure`

---

* `benchmark_runner/workloads/workloads_operations.py: remove trailing spaces`

---

PR #393, opened 2022-03-16
template_operations: add end-of-line at end-of-generated-file

kpouget:
This PR adds the EOF character at the end of template-generated files (following UNIX best practises)
and updates the files in `tests/unittest/benchmark_runner/common/template_operations/golden_files` accordingly.


benchmark_runner/common/template_operations/template_operations.py
```diff
@@ -120,7 +120,7 @@ def get_sub_dict(dictionary: dict, key: str, value: str):
                 template_file = f'{file_components[0]}_template{file_components[1]}'
             with open(os.path.join(workload_dir_path, 'internal_data', template_file)) as f:
                 template = Template(f.read())
-            answer[filename] = template.render(render_data)
+            answer[filename] = template.render(render_data) + "\n"
```
ebattat, line 123: Please use formatting instead of + (generate new objects in memory)
answer[filename] = f'{template.render(render_data)}\n'

kpouget: I don't think that the f-string and the `+` are much different in memory (considering that we're adding a single-char),
and I find the intend clearer with the concatenation (add EOL at EOF),
but I can change it to the f-string if you prefer

kpouget:
> I notice the golden file check on #391 is failing. Is that because this is supposed to fix it? If so, please roll this into #391. The golden file should never get out of sync.

yes, I'll remove this commit (the `+ \n`) from the other PR, so that golden file changes (144 files) are not mixed with the rest of the functional changes

RobertKrawitz (approved):
lgtm

kpouget:
> That's good practice, also because mixing consequential changes with trivial changes such as this risks masking changes that matter.

exactly, and that's why I keep my commits grouped logically, and ideally don't want them squashed, but I feel like Github-like tools made that practise disappear :#

benchmark_runner/common/template_operations/template_operations.py
kpouget: I found what really bothered be with the f-string and that I wouldn't written in my code:
```
template.render(render_data) + "\n"
```
vs 
```
f"{template.render(render_data)}\n"
```
`template.render(render_data)` is a complex call (no side effect, but still not a trivial function), 
and it gets mixed with a formatting operation.

This would make more sense to me:
```
rendered_template = template.render(render_data)
answer = f"{rendered_template}\n"
```

PR #416, opened 2022-04-20
Migrate infra to non-worker nodes

RobertKrawitz:
Not yet tested.


benchmark_runner/common/ocp_resources/migrate_infra.py
```diff
@@ -0,0 +1,35 @@
...
+    def create_odf(self):
+        """
+        This method create odf
+        :return:
+        """
+        for resource in self.__resource_list:
+            logger.info(f'run {resource}')
+            if resource == '01_cluster-monitoring-configmap-template.yaml':
+                nodes = self.__oc.get_master_nodes().split()
+                if len(nodes) < 1:
```
portante, line 30: What is the purpose of this test?

RobertKrawitz: If we don't have any infra nodes, we don't want to try to bind the infra components to infra nodes.

benchmark_runner/common/ocp_resources/migrate_infra.py
```diff
@@ -27,9 +26,8 @@ def create_odf(self):
             logger.info(f'run {resource}')
             if resource == '01_cluster-monitoring-configmap-template.yaml':
                 nodes = self.__oc.get_master_nodes().split()
-                if len(nodes) < 1:
-                    pass
-                for node in nodes:
-                    self.run(f'oc label {node} node-role.kubernetes.io/infra=')
-                self.__oc._create_async(yaml=os.path.join(self.__path, resource))
+                if len(nodes) >= 1:
```
portante, line 29: FWIW, since `.split()` returns a `list`, `if nodes:` is sufficient.

PR #418, opened 2022-04-23
enable infra

ebattat:


ebattat:
Enable Infra in Operator installation

PR #425, opened 2022-05-10
Basic support for running workloads in kubernetes cluster

gqlo:
- stress-ng-pod working
- vdbench-pod working
- detailed check is needed to address differences in kubectl with k8s

### DIFF
--- a/benchmark_runner/common/oc/oc.py
+++ b/benchmark_runner/common/oc/oc.py
@@ -27,47 +27,49 @@ class OC(SSH):
         self.__environment_variables_dict = environment_variables.environment_variables_dict
         self.__run_artifacts = self.__environment_variables_dict.get('run_artifacts_path', '')
         self.__elasticsearch_url = self.__environment_variables_dict.get('elasticsearch_url', '')
+        self.__cli = self.__environment_variables_dict.get('cli', '')
 
     def get_ocp_server_version(self):
         """
         This method return ocp server version
         :return:
         """
-        return self.run("oc version -ojson | jq -r '.openshiftVersion'")
+        return self.run(f"{self.__cli} version -ojson | jq -r '.openshiftVersion'")
 
     def get_cnv_version(self):
         """
         This method return cnv version
         :return:
         """
-        return self.run("oc get csv -n openshift-cnv $(oc get csv -n openshift-cnv --no-headers | awk '{ print $1; }') -ojsonpath='{.spec.version}'")
+        return self.run(f"{self.__cli} get csv -n openshift-cnv $(oc get csv -n openshift-cnv --no-headers | awk '{{ print $1; }}') -ojsonpath='{{.spec.version}}'")
 
     def get_odf_version(self):
         """
         This method return odf version
         :return:
         """
-        return self.run("oc get csv -n openshift-storage -ojsonpath='{.items[0].spec.labels.full_version}'")
+        return self.run(f"{self.__cli} get csv -n openshift-storage "
+                        f"-ojsonpath='{{.items[0].spec.labels.full_version}}'")
 
     def get_kata_version(self):
         """
         This method returns kata version
         :return:
         """
-        return self.run("oc get csv -n openshift-sandboxed-containers-operator $(oc get csv -n openshift-sandboxed-containers-operator --no-headers | awk '{ print $1; }') -ojsonpath='{.spec.version}' ")
+        return self.run(f"{self.__cli} get csv -n openshift-sandboxed-containers-operator $({self.__cli} get csv -n openshift-sandboxed-containers-operator --no-headers | awk '{{ print $1; }}') -ojsonpath='{{.spec.version}}'")
 
     def _get_kata_default_channel(self):
         """
         Retrieve the default channel for Kata
         """
-        return self.run("oc get packagemanifest -n openshift-marketplace sandboxed-containers-operator -ojsonpath='{.status.defaultChannel}'")
+        return self.run(f"{self.__cli} get packagemanifest -n openshift-marketplace sandboxed-containers-operator -ojsonpath='{{.status.defaultChannel}}'")
 
     def _get_kata_default_channel_field(self, channel_field: str):
         """
         Retrieve a field from the packagemanifest for the default Kata channel
         """
         default_channel = f'"{self._get_kata_default_channel()}"'
-        command=f"oc get packagemanifest -n openshift-marketplace sandboxed-containers-operator -ojson | jq -r '[foreach .status.channels[] as $channel ([[],[]];0; (if ($channel.name == {default_channel}) then $channel.{channel_field} else null end))] | flatten | map (select (. != null))[]'"
+        command = f"{self.__cli} get packagemanifest -n openshift-marketplace sandboxed-containers-operator -ojson | jq -r '[foreach .status.channels[] as $channel ([[],[]];0;(if ($channel.name == {default_channel}) then $channel.{channel_field} else null end))] | flatten | map (select (. != null))[]'"
         return self.run(command)
 
     def _get_kata_csv(self):
@@ -80,7 +82,7 @@ class OC(SSH):
         """
         Retrieve the catalog source of the sandboxed containers operator for installation"
         """
-        return self.run("oc get packagemanifest -n openshift-marketplace sandboxed-containers-operator -ojsonpath='{.status.catalogSource}'")
+        return self.run(f"{self.__cli} get packagemanifest -n openshift-marketplace sandboxed-containers-operator -ojsonpath='{{.status.catalogSource}}'")
 
     def _get_kata_channel(self):
         """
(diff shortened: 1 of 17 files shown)


benchmark_runner/common/template_operations/render_yaml_from_template.py
```diff
@@ -16,12 +16,14 @@ def render_yaml_string(template_str: str, environment_variable_dict: dict = None
     if not environment_variable_dict:
         environment_variable_dict = environment_variables.environment_variables_dict
     # Allow faking the Prometheus token so that we can generate deterministic test files
-    if 'prom_token_override' in environment_variable_dict:
-        environment_variable_dict['prom_token'] = environment_variable_dict.get('prom_token_override', '')
-    else:
-        oc = OC(kubeadmin_password=environment_variable_dict.get('kubeadmin_password', ''))
-        # set prom token
-        environment_variable_dict['prom_token'] = oc.get_prom_token()
+    if environment_variable_dict.get("enable_prometheus_snapshot", None):
```
ebattat, line 19: Just run:
if environment_variable_dict.get("enable_prometheus_snapshot"):
No need None

benchmark_runner/main/environment_variables.py
```diff
@@ -37,6 +37,8 @@ def __init__(self):
         self._environment_variables_dict['runner_path'] = os.environ.get('RUNNER_PATH', '/tmp')
         # This path is for vm/pod/prometheus run artifacts
         self._environment_variables_dict['run_artifacts'] = os.environ.get('RUN_ARTIFACTS', os.path.join(self._environment_variables_dict['runner_path'], 'benchmark-runner-run-artifacts'))
+        # openshift or kubernetes
+        self._environment_variables_dict['cluster'] = os.environ.get('CLUSTER', 'oc')
```
ebattat, line 41: default should be 'openshift' instead of 'oc'

benchmark_runner/common/oc/oc.py
```diff
@@ -293,11 +302,11 @@ def get_pod(self, label: str, database: str = ''):
         """
         if database:
             return self.run(
-                f"oc get pods -n '{database}-db'" + " --no-headers | awk '{ print $1; }' | grep " + database,
+                f"{self.__cli} get pods -n '{database}-db'" + " --no-headers | awk '{ print $1; }' | grep " + database,
                 is_check=True).rstrip().decode('ascii')
         else:
             return self.run(
-                "oc get pods -n " + environment_variables.environment_variables_dict['namespace'] + " --no-headers | awk '{ print $1; }' | grep -w " + label,
+                f"{self.__cli} get pods -n " + environment_variables.environment_variables_dict['namespace'] + " --no-headers | awk '{ print $1; }' | grep -w " + label,
```
RobertKrawitz, line 309:
```
f"{self.__cli} get pods -n {environment_variables.environment_variables_dict['namespace']}  --no-headers | awk '{{ print $1; }}' | grep -w '{label}'"
'''
Reference: https://www.python.org/dev/peps/pep-0498/#specification

File: benchmark_runner/common/oc/oc.py [hunk 4]
```diff
@@ -310,10 +319,10 @@ def get_vm(self, label: str = '', namespace: str = environment_variables.environ
         """
         if label:
             return self.run(
-                cmd="oc get vmi -n " + namespace + " --no-headers | awk '{ print $1; }' | grep -w " + label,
+                cmd=f"o{self.__cli}c get vmi -n " + namespace + " --no-headers | awk '{ print $1; }' | grep -w " + label,
```
RobertKrawitz, line 322: Same: do it as a single f-string.

RobertKrawitz (changes requested):
Minor changes.

benchmark_runner/common/template_operations/render_yaml_from_template.py
RobertKrawitz: It's not necessary, because `dict.get()` defaults the second argument to None.

benchmark_runner/common/oc/oc.py
```diff
@@ -27,47 +27,55 @@ def __init__(self, kubeadmin_password: str = ''):
...
+                        f"-ojsonpath='{{.status.defaultChannel}}'")
 
     def _get_kata_default_channel_field(self, channel_field: str):
         """
         Retrieve a field from the packagemanifest for the default Kata channel
         """
         default_channel = f'"{self._get_kata_default_channel()}"'
-        command=f"oc get packagemanifest -n openshift-marketplace sandboxed-containers-operator -ojson | jq -r '[foreach .status.channels[] as $channel ([[],[]];0; (if ($channel.name == {default_channel}) then $channel.{channel_field} else null end))] | flatten | map (select (. != null))[]'"
+        command = f"{self.__cli} get packagemanifest -n openshift-marketplace sandboxed-containers-operator " \
+                  f"-ojson | jq -r '[foreach .status.channels[] as $channel ([[],[]];0; " \
```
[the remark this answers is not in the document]
gqlo: Hi Robert, using with `f"""` will cause the command span multiple lines (unless we keep the command a long single line), It probably won't work. Have you tried to print out the string?

benchmark_runner/common/oc/oc.py
gqlo: I changed it to f string, what is the purpose of variable `cmd` here?

benchmark_runner/common/template_operations/render_yaml_from_template.py
gqlo: That's right, just removed the None.

benchmark_runner/common/oc/oc.py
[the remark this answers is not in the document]
gqlo: Anyway, I just made all commands into a single line for consistency.

benchmark_runner/main/environment_variables.py
```diff
@@ -37,6 +37,8 @@ def __init__(self):
         self._environment_variables_dict['runner_path'] = os.environ.get('RUNNER_PATH', '/tmp')
         # This path is for vm/pod/prometheus run artifacts
         self._environment_variables_dict['run_artifacts'] = os.environ.get('RUN_ARTIFACTS', os.path.join(self._environment_variables_dict['runner_path'], 'benchmark-runner-run-artifacts'))
+        # openshift or kubernetes
```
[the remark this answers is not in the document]
gqlo: thx

benchmark_runner/common/oc/oc.py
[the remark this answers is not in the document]
RobertKrawitz: You could use `\` at the end of the line if needed, but this is OK.

benchmark_runner/common/oc/oc.py
RobertKrawitz: It's treating it as a keyword although it's a positional parameter.  It doesn't do any harm, but it shouldn't really be needed here.

benchmark_runner/common/template_operations/render_yaml_from_template.py
```diff
@@ -20,8 +20,9 @@ def render_yaml_string(template_str: str, environment_variable_dict: dict = None
         environment_variable_dict['prom_token'] = environment_variable_dict.get('prom_token_override', '')
     else:
         oc = OC(kubeadmin_password=environment_variable_dict.get('kubeadmin_password', ''))
-        # set prom token
-        environment_variable_dict['prom_token'] = oc.get_prom_token()
+        # set prom token, skip if running on kubernetes just for now
+        if environment_variable_dict.get("cluster") != 'kubernetes':
```
ebattat, line 24: @gqlo, Why did you ask for negative, why not positive:
if environment_variable_dict.get("cluster") == 'openshift':

gqlo: The reason is that I wanna make it clear we are doing a dirty hack for kubernetes, if we are using "openshift" then by reading this line, we won't be able to understand this line is changed for the purpose of skipping prom token in kubernetes.

ebattat: I think that asking for positive is more clear and more readability

gqlo: `if environment_variable_dict.get("cluster") == 'openshift':`  By reading this line, as you can see nothing tells us about kubernetes. let's say if we have a new teammate come to our team, it would be very difficult for him to guess this line has something to do with k8s. I tend to stick to this principle as much as possible so that new teammates needs less context to understand the code. It probably won't make much difference for now, since we might need to remove the whole condition in the future if we want to run workload with prometheus in k8s.

I would be interested to know why you think a positive check is more clear and readable. Can you elaborate with more details?

ebattat: For me its more clear, we can leave it as is.

ebattat (approved):
/LGTM

RobertKrawitz (approved):
/lgtm

gqlo:
Added template to run vdbench-pod on k8s. Another round of review is needed before merge.

ebattat (approved):
/LGTM

PR #428, opened 2022-05-16
Add prometheus-api-client and openshift-client to support clusterbuster

RobertKrawitz:
Packages required for ClusterBuster-based runs.

### DIFF
--- a/tests_requirements.txt
+++ b/tests_requirements.txt
@@ -4,7 +4,9 @@ elasticsearch_dsl==7.4.0
 jinja2==3.0.3
 mock
 moto==2.2.16
+openshift-client==1.0.14
 pandas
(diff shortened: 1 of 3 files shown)


tests_requirements.txt
```diff
@@ -4,7 +4,9 @@ elasticsearch_dsl==7.4.0
 jinja2==3.0.3
 mock
 moto==2.2.16
+openshift-client==1.0.14
 pandas
+prometheus-api-client==0.5.0
```
ebattat, line 9: Please add it to Setup.py instead of  tests_requirements.txt
tests_requirements.txt is only for tests

RobertKrawitz: Done.

PR #429, opened 2022-05-19
set the optimal number of threads per tests

bbenshab:
I have set the optimal number of threads per test in order to get repetitive consistent results on the perf ci for the best possible regression testing.


benchmark_runner/common/template_operations/templates/vdbench/internal_data/vdbench_pod_template.yaml
```diff
@@ -63,6 +63,10 @@ metadata:
   name: vdbench-{{ kind }}-{{ trunc_uuid }}-{{ scale }}
   {%- endif %}
   namespace: {{ namespace }}
+  annotations:
```
ebattat, line 66: Please put if before annotations

RobertKrawitz: I agree; an empty annotations list does nothing.

However, it might not be a bad idea to put this in a data template in a way that can be reused with other workloads.  j2 being j2, though, that's not always very simple.

benchmark_runner/common/template_operations/templates/vdbench/internal_data/vdbench_pod_template.yaml
```diff
@@ -63,6 +63,10 @@ metadata:
...
+    {%- if kind == 'kata' %}
+    io.katacontainers.config.hypervisor.virtio_fs_extra_args: '["-o","allow_direct_io"]'
```
RobertKrawitz, line 68: Do be certain to test this to ensure that it does what you expect.

bbenshab: Yeah, I tested it.

ebattat (approved):
/LGTM

benchmark_runner/common/template_operations/templates/vdbench/internal_data/vdbench_pod_template.yaml
bbenshab: @RobertKrawitz  I'm not sure exactly what you are suggesting here, but IMHO the "annotation" key-value should be in any pod/VM YAML it's the correct way, same as the "label" key-value for convenience and ease of use, especially when doing some manual intervention for investigations.

PR #432, opened 2022-06-03
update perf timeout to 2 hours

ebattat:
This PR is for rebuilding benchmark-operator inside benchmark-runner


ebattat:
No need to update it for now

PR #441, opened 2022-07-03
ClusterBuster: save run artifacts and upload to s3

ebattat:
Add to clusterbuster:

1. Save local artifacts and tar it
2. Upload results to s3
3. Add Prometheus


ebattat (commented):
/LTGM

PR #449, opened 2022-07-20
disable Prometheus snapshot when ENABLE_PROMETHEUS_SNAPSHOT is False

ebattat:
This PR include:
1. Fix Prometheus enable 
2. Split run_artifacts from upload s3
3. Fix func and perf workflows


benchmark_runner/benchmark_operator/benchmark_operator_workloads_operations.py
```diff
@@ -470,16 +478,18 @@ def initialize_workload(self):
...
         """
         This method includes all the finalization of workload
         :return:
         """
-        if self._enable_prometheus_snapshot:
+        if self._enable_prometheus_snapshot == 'True':
             self.end_prometheus()
         if self._endpoint_url:
             self.upload_run_artifacts_to_s3()
+        if self._save_artifacts_local == 'False':
```
RobertKrawitz, line 493: Compare everything against "True" or "False", not some things against one and some things against the other (what happens if the value is something different?).

RobertKrawitz (changes requested):
Please retitle the PR

Correctly disable Prometheus snapshot when ENABLE_PROMETHEUS_SNAPSHOT is False

for clarity.

benchmark_runner/benchmark_operator/benchmark_operator_workloads_operations.py
RobertKrawitz: Not needed.

RobertKrawitz (approved):
lgtm

PR #450, opened 2022-07-20
Handle boolean options as bools rather than as strings

RobertKrawitz:
Boolean options are currently handled as strings and compared textually against "True" and "False".  They should be parsed and then treated as bools.

### DIFF
--- a/benchmark_runner/common/template_operations/template_operations.py
+++ b/benchmark_runner/common/template_operations/template_operations.py
@@ -32,7 +32,7 @@ class TemplateOperations:
         Kata shares templates with pod
         """
         if self.__workload_kind == 'kata' or self.__workload_kind == 'pod':
-            if self.__environment_variables_dict.get('kata_cpuoffline_workaround', '') == 'True':
+            if self.__environment_variables_dict.get('kata_cpuoffline_workaround', False):
                 logger.warn('*** WARNING: Enabling Kata CPU offline workaround ***')
             return 'pod'
         elif self.__workload_kind == 'vm':
@@ -123,7 +123,7 @@ class TemplateOperations:
 
         workload_data = yaml.load(render_yaml_file(workload_dir_path, f'{self.__workload_name}_data_template.yaml', template_render_data), Loader=yaml.FullLoader)
 
-        if self.__environment_variables_dict.get("config_from_args") == "True":
+        if self.__environment_variables_dict.get("config_from_args"):
             config = dict([kv.split("=") for kv in sys.argv[1:]])
             workload_data["template_data"]["run_type"]["default"] = config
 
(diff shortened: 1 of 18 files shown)


RobertKrawitz:
This will require full e2e testing.

ebattat (commented):
@RobertKrawitz, Did you test it ?

benchmark_runner/workloads/workloads_operations.py
```diff
@@ -29,7 +29,7 @@ def __init__(self):
         self._workload = self._environment_variables_dict.get('workload', '')
         self._build_version = self._environment_variables_dict.get('build_version', '')
         self._workloads_odf_pvc = list(self._environment_variables_dict.get('workloads_odf_pvc', ''))
-        self._odf_pvc = self._environment_variables_dict.get('odf_pvc', '')
+        self._odf_pvc = self._environment_variables_dict.get('odf_pvc', True)
```
gqlo, line 32: By default is True?

benchmark_runner/workloads/workloads_operations.py
```diff
@@ -40,7 +40,7 @@ def __init__(self):
         self._endpoint_url = self._environment_variables_dict.get('endpoint_url', '')
         self._run_artifacts_path = self._environment_variables_dict.get('run_artifacts_path', '')
         self._save_artifacts_local = self._environment_variables_dict.get('save_artifacts_local', '')
-        self._enable_prometheus_snapshot = self._environment_variables_dict.get('enable_prometheus_snapshot', '')
+        self._enable_prometheus_snapshot = self._environment_variables_dict.get('enable_prometheus_snapshot', True)
```
gqlo, line 43: Same here

benchmark_runner/workloads/workloads_operations.py
RobertKrawitz: @ebattat I believe this is the current default?

benchmark_runner/workloads/workloads_operations.py
RobertKrawitz: @ebattat is this what's intended?

benchmark_runner/workloads/workloads_operations.py
ebattat: Yes, True in the current default

benchmark_runner/workloads/workloads_operations.py
ebattat: False is the default

RobertKrawitz: Fixed.

benchmark_runner/common/template_operations/template_operations.py
```diff
@@ -32,7 +32,7 @@ def __get_workload_template_kind(self):
         Kata shares templates with pod
         """
         if self.__workload_kind == 'kata' or self.__workload_kind == 'pod':
-            if self.__environment_variables_dict.get('kata_cpuoffline_workaround', '') == 'True':
+            if self.__environment_variables_dict.get('kata_cpuoffline_workaround', False):
```
ebattat, line 35: Default should be True, Am I right ?

ebattat (approved):

PR #452, opened 2022-07-21
Correct the command to generate a Prometheus token

RobertKrawitz:
oc sa new-token is deprecated; this change is needed so that the command doesn't generate spurious stderr that pollutes the output.


ebattat (approved):
/LGTM

RobertKrawitz:
Merged into PR #450

PR #456, opened 2022-07-26
add clusterbuster termination timeout

ebattat:
Add clusterbuster force termination timeout in case of stuck during termination

That should solve the following issue:

```
# oc get pod -A -l clusterbuster
NAMESPACE          NAME                             READY   STATUS        RESTARTS   AGE
clusterbuster-10   clusterbuster-10-cpusoaker-5-1   0/1     Terminating   0          6m18s
```


RobertKrawitz (approved):
lgtm.

There's no guarantee that this will work reliably.  If a pod fails to terminate cleanly, it's possible that something will be left around that will cause problems in the future, as discussed here: https://access.redhat.com/solutions/4165791

This is an attempt to get around the problem for CI.

PR #470, opened 2022-09-15
add clusterbuster release workflow

ebattat:
This workflow will run clusterbuster release run (~16 hours) against Perf environment
The workflow will triggered manually


RobertKrawitz (approved):
The code's rather complex; unfortunate that we can't test it.  Just one tiny comment nit that won't stop me from approving it.

PR #472, opened 2022-09-21
fix kata virtiofsd permission issue

ebattat:
This PR should supply temporary workaround for kata virtiofsd permission issue 
**The issue:**
kata database failed due to permission issue in mssql/mariadb/pssql: cannot create directory '/var/lib/mysql/data': Permission denied

**Root cause:**
Failed only for kata using ODF4.10.5, working for kata using ODF4.9.10 (regardless of whether OCP 4.10 or 4.11 is in use).  virtiofsd does not check secondary group ids; it only relies on the primary group id.

**The temporary workaround:**
Add to the deployment spec:
```

securityContext:
  fsGroup: 0

And the SCC:

apiVersion: security.openshift.io/v1
metadata:
  name: mariadbscc
kind: SecurityContextConstraints
fsGroup:
  type: RunAsAny
runAsUser:
  type: MustRunAsRange
seLinuxContext:
  type: MustRunAs
users:
  - system:serviceaccount:mariadb-db:default

```

### DIFF
--- a/benchmark_runner/common/template_operations/templates/hammerdb/internal_data/mariadb_template.yaml
+++ b/benchmark_runner/common/template_operations/templates/hammerdb/internal_data/mariadb_template.yaml
@@ -3,6 +3,23 @@ kind: Namespace
 metadata:
   name: mariadb-db
 ---
+{%- if kind == 'kata' %}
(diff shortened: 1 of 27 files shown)


RobertKrawitz (approved):
lgtm

PR #474, opened 2022-10-03
merge FUNC and PERF env variables

ebattat:
This PR is merged FUNC and PERF env variable.
No reason to hold those 2 prefix for the environment type
only the secret determine if it FUNC and PERF environment


RobertKrawitz (approved):
lgtm

PR #485, opened 2022-11-02
fix PostgreSQL LSO

ebattat:
This PR should fix PostgreSQL LSO hammerdb issue.
The current results of ODF are better than LSO because lack of load transactions.

There are 2 changes in this PR:
1. Update load transactions to 200K instead of 100K
2. Add LSO verification in case of empty path.

### DIFF
--- a/benchmark_runner/benchmark_operator/benchmark_operator_workloads_operations.py
+++ b/benchmark_runner/benchmark_operator/benchmark_operator_workloads_operations.py
@@ -11,7 +11,7 @@ from benchmark_runner.common.oc.oc import OC
 from benchmark_runner.common.template_operations.template_operations import TemplateOperations
 from benchmark_runner.common.elasticsearch.elasticsearch_operations import ElasticSearchOperations
 from benchmark_runner.common.ssh.ssh import SSH
-from benchmark_runner.benchmark_operator.benchmark_operator_exceptions import ODFNonInstalled
+from benchmark_runner.benchmark_operator.benchmark_operator_exceptions import ODFNonInstalled, EmptyLSOPath
 from benchmark_runner.main.environment_variables import environment_variables
 from benchmark_runner.common.clouds.shared.s3.s3_operations import S3Operations
 from benchmark_runner.common.prometheus.prometheus_snapshot import PrometheusSnapshot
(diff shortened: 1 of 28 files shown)


benchmark_runner/benchmark_operator/benchmark_operator_workloads_operations.py
```diff
@@ -455,6 +456,15 @@ def odf_pvc_verification(self):
             if not self._oc.is_odf_installed():
                 raise ODFNonInstalled()
 
+    @logger_time_stamp
+    def lso_verification(self):
+        """
+        This method verify that lso path exist
```
RobertKrawitz, line 462: Verifies

benchmark_runner/benchmark_operator/benchmark_operator_workloads_operations.py
```diff
@@ -455,6 +456,15 @@ def odf_pvc_verification(self):
             if not self._oc.is_odf_installed():
                 raise ODFNonInstalled()
 
+    @logger_time_stamp
+    def lso_verification(self):
```
RobertKrawitz, line 460: "verify_lso" would be a better name.

benchmark_runner/common/template_operations/templates/hammerdb/hammerdb_data_template.yaml
```diff
@@ -8,7 +8,6 @@ template_data:
   shared:
     pin_node1: {{ pin_node1 }}
     pin_node2: {{ pin_node2 }}
-    transactions: 100000
```
RobertKrawitz, line 11: You should be able to leave this here, and override it only for postgres.  You can also put the rampup time here.

RobertKrawitz (changes requested):

benchmark_runner/benchmark_operator/benchmark_operator_workloads_operations.py
ebattat: Fixed

benchmark_runner/common/template_operations/templates/hammerdb/hammerdb_data_template.yaml
ebattat: Fixed

RobertKrawitz (approved):
Noted in private discussion that the change of runtime from 2 to 1 is intentional.  Approving.

PR #492, opened 2022-11-21
remove coredns patch

ebattat:
Remove coredns patch because removing bufsize from /etc/coredns/Corefile


ebattat:
After discuss it with @RobertKrawitz, understand that its no harm to leave it in case of installing OCP 4.10 or older versions.

jeniferh:
Sounds good to leave as is, and glad to see the issue is resolved in newer OCP versions.

ebattat:
@RobertKrawitz, do you think we need to close this PR or leave it as is ?

ebattat:
Close due to maintenance

PR #493, opened 2022-11-21
update ocp version and Grafana dashboards

ebattat:
1. Update ocp version from upstream URL.
2. Update latest grafana dashboards

### DIFF
--- a/benchmark_runner/common/clouds/IBM/ibm_operations.py
+++ b/benchmark_runner/common/clouds/IBM/ibm_operations.py
@@ -29,7 +29,7 @@ class IBMOperations:
     """
     This class is responsible for all IBM cloud operations, all commands run on remote provision IBM host
     """
-    LATEST_VERSION = 'latest'
+    LATEST = 'latest'
 
     @typechecked
     def __init__(self, user: str):
@@ -182,18 +182,17 @@ class IBMOperations:
         """
         self.__remote_ssh.disconnect()
 
-    def __get_latest_version(self):
+    def __get_installation_version(self):
         """
-        This method return latest version
+        This method return installation version
         :return:
         """
         ocp_versions = OCPVersions()
-        if self.LATEST_VERSION in self.__install_ocp_version:
+        if self.LATEST in self.__install_ocp_version:
             openshift_version_data = self.__install_ocp_version.split('-')
             return ocp_versions.get_latest_version(latest_version=openshift_version_data[1])
         else:
-            openshift_version_data = self.__install_ocp_version.split('.')
-            return ocp_versions.get_latest_version(latest_version=f'{openshift_version_data[0]}.{openshift_version_data[1]}')
+            return self.__install_ocp_version
 
     def get_ocp_server_version(self):
         """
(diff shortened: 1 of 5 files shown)


ebattat:
We are fetching the OCP version from the cluster before having the new cluster credentials, so solve it by fetching it from assisted installer upstream url.

RobertKrawitz:
Looking at `OCPVersions.get_latest_version`, it looks like this simply gets the latest available version, not necessarily the one we're installing.

ebattat:
@RobertKrawitz, it preferred way to show which version installed instead of not showing at all.

RobertKrawitz:
What happens if we deliberately install a version other than the very latest version available (e. g. we choose to install 4.11 even after 4.12 is released)?

ebattat:
@RobertKrawitz, you right fix it. in case of passing an constant version, I will take it as it, meaning the version you pass from [INSTALL_OCP_VERSION environment variable](https://github.com/redhat-performance/benchmark-runner/blob/ea82ed22997cc0582653e4493a242857234760b3/.github/workflows/Weekly_Perf_Env_Installer_CI.yml#L67).

ebattat:
@RobertKrawitz agreed, fixed.

benchmark_runner/common/clouds/IBM/ibm_operations.py
```diff
@@ -209,7 +208,7 @@ def version_already_installed(self):
         """
         if self.get_ocp_server_version() == 'null':
             return False
-        elif self.LATEST_VERSION in self.__install_ocp_version and self.get_ocp_server_version() == self.__get_latest_version():
+        elif self.LATEST_VERSION in self.__install_ocp_version and self.get_ocp_server_version() == self.__get_installation_version():
```
RobertKrawitz, line 211: LATEST_VERSION appears to be the fixed string `latest`.  I'm not quite sure how that's working here?

ebattat: What do you mean, LATEST instead of LATEST_VERSION ?

RobertKrawitz: That wasn't what I meant, and no need to change it.  I'm trying to work out what happens if the version specified is `latest`, how that gets translated into a specific version.

ebattat: This is already working. Please let me know if you have more comments.

RobertKrawitz: I'll approve it as I don't believe you have made any material change, but I would like to understand how specifying `latest` works.

RobertKrawitz: Before I approve it -- I notice that the current diff has a lot of changes to various JSON files.  What's that about?

ebattat: I have update our grafana dashboards, adding it into description.

RobertKrawitz: That ideally should be a separate PR, but I'll OK this.  I've updated the title too.

RobertKrawitz (approved):

PR #501, opened 2022-12-15
split bootstorm vm

ebattat:
#Fixes

Split start vm from wait for login and run it on separate threads


ebattat:
We cannot split start from wait, all must be on the same method because we calculate start time till login

PR #502, opened 2022-12-15
PostgreSQL VM autovacuum off

ebattat:
Fixes:

autovacuum off in PostgreSQL VM, depend on benchmark-operator [PR](https://github.com/cloud-bulldozer/benchmark-operator/pull/801)

git clone -b v1.0.2 https://github.com/cloud-bulldozer/benchmark-operator =>  [autovacuum off](https://github.com/cloud-bulldozer/benchmark-operator/blob/095443730bd9a5e5fb0bd128fe18119a7cc8dfda/roles/hammerdb/templates/db_postgres_workload_vm.sh.j2#L61) in PostgreSQL VM

### DIFF
--- a/.github/workflows/Func_Env_Build_Test_CI.yml
+++ b/.github/workflows/Func_Env_Build_Test_CI.yml
@@ -77,8 +77,8 @@ jobs:
           cp $RUNNER_PATH/kubectl /usr/local/bin/kubectl
           cp $RUNNER_PATH/oc /usr/local/bin/oc
           
-          # clone benchmark-operator v1.0.1
-          git clone -b v1.0.1 https://github.com/cloud-bulldozer/benchmark-operator $RUNNER_PATH/benchmark-operator
+          # clone benchmark-operator v1.0.2
+          git clone -b v1.0.2 https://github.com/cloud-bulldozer/benchmark-operator $RUNNER_PATH/benchmark-operator
           
           # run pytest - only unittest
           pytest -v tests/unittest/
@@ -155,7 +155,7 @@ jobs:
         rm -rf $RUNNER_PATH/virtctl
         
         # clone benchmark-operator
-        git clone https://github.com/cloud-bulldozer/benchmark-operator $RUNNER_PATH/benchmark-operator
+        git clone -b v1.0.2 https://github.com/cloud-bulldozer/benchmark-operator $RUNNER_PATH/benchmark-operator
         # run pytest
         pytest -v tests --cov=benchmark_runner --cov-report=term-missing
         coverage run -m pytest -v tests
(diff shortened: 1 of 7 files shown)


RobertKrawitz (changes requested):
Noting that benchmark-operator v1.0.2 contains the needed fix.

- Various workflow files still reference either v1.0.1 or no version at all.
- HOW_TO, README, docs/source/develop.md, and docs/source.podman.md also do not reference a version.
- benchmark_runner/common/template_operations/templates/common.yaml should either reference a fixed version of metrics.yaml or that file should be carried here.

`git grep cloud-bulldozer/benchmark-operator` should help find what needs to be fixed.

ebattat:
@RobertKrawitz 
See my answer below:

Various workflow files still reference either v1.0.1 or no version at all.
=> I dont think so because this tag generated for us.

HOW_TO, README, docs/source/develop.md, and docs/source.podman.md also do not reference a version.
=> no need add reference to benchmark-operator version because its integrated inside benchmark-runner container, same as we dont need to mention clusterbuster version.

benchmark_runner/common/template_operations/templates/common.yaml should either reference a fixed version of metrics.yaml or that file should be carried here.
=> I think it should be on separate PR

RobertKrawitz:
> @RobertKrawitz See my answer below:
> 
> Various workflow files still reference either v1.0.1 or no version at all. => All workflows are using the same version, BTW v1.0.2 version generate to us.

Why should that be referencing a different version, then?

> HOW_TO, README, docs/source/develop.md, and docs/source.podman.md also do not reference a version. => no need add reference to benchmark-operator version because its integrated inside benchmark-runner container, same as we dont need to mention clusterbuster version.

I think it should reference the same version; if there are incompatible changes in benchmark-operator, that might cause difficulties.

> benchmark_runner/common/template_operations/templates/common.yaml should either reference a fixed version of metrics.yaml or that file should be carried here. => I think it should be on separate PR

That's fine.

ebattat:
@RobertKrawitz,

Various workflow files still reference either v1.0.1 or no version at all. => All workflows are using the same version, BTW v1.0.2 version generate to us.

Why should that be referencing a different version, then?

**I have pushed autovacuum = off fix into benchmark-operator and then ask to tag it, see [PR](https://github.com/cloud-bulldozer/benchmark-operator/pull/801)**

HOW_TO, README, docs/source/develop.md, and docs/source.podman.md also do not reference a version. => no need add reference to benchmark-operator version because its integrated inside benchmark-runner container, same as we dont need to mention clusterbuster version.

I think it should reference the same version; if there are incompatible changes in benchmark-operator, that might cause difficulties.

**The best practice is to package it and add it to the dependency in setup.py, but if there is no option to the solution is to containerized it.**

benchmark_runner/common/template_operations/templates/common.yaml should either reference a fixed version of metrics.yaml or that file should be carried here. => I think it should be on separate PR

That's fine.

RobertKrawitz (changes requested):
The doc (HOW_TO.md, README.md, docs/source/develop.md, docs/source/podman.md) should specify the explicit version, in case further development in benchmark-operator breaks us.  Otherwise, I believe it's set.

RobertKrawitz (approved):

PR #508, opened 2022-12-19
fix scale issue

ebattat:
Fix scale parameter issue by remove int casting

### DIFF
--- a/benchmark_runner/common/template_operations/template_operations.py
+++ b/benchmark_runner/common/template_operations/template_operations.py
@@ -124,7 +124,7 @@ class TemplateOperations:
             'workload_extra_name': self.__workload_extra_name,
             'standard_output_file': self.__standard_output_file,
             'standard_template_file': self.__standard_template_file,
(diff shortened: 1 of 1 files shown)


benchmark_runner/common/template_operations/template_operations.py
```diff
@@ -124,7 +124,7 @@ def generate_yamls_internal(self, scale: str = None, scale_node: str = None, red
             'workload_extra_name': self.__workload_extra_name,
             'standard_output_file': self.__standard_output_file,
             'standard_template_file': self.__standard_template_file,
-            'scale': int(scale)+1,
+            'scale': scale,
```
RobertKrawitz, line 127: Why isn't this
```
str(int(scale)+1)
```
?

ebattat: We cannot use int just string, revert it back to scale

RobertKrawitz (approved):
This is in fact correct.

PR #509, opened 2022-12-20
Multi-threads: add scale number

ebattat:
Fixes:

Create a separate render variable to scale number for update it in state_signals_exporter_pod.yaml


benchmark_runner/common/template_operations/template_operations.py
```diff
@@ -94,12 +94,15 @@ def __build_template_data(self, previous_data: dict, template_root_data: dict):
                 **extra_data, **extra_kind_data, **extra_runtype_data, **extra_kind_runtype_data}
 
     @logger_time_stamp
-    def generate_yamls_internal(self, scale: str = None, scale_node: str = None, redis: str = None):
+    def generate_yamls_internal(self, scale: str = None, scale_num: str = None, scale_node: str = None, redis: str = None):
```
RobertKrawitz, line 97: It looks like `scale` here is really an index, and `scale_num` is the total scale.  I suggest naming these parameters `scale_idx` and `scale` or similar.

These should be passed in as ints, and this routine should convert them to strings.  These are conceptually numbers, but they need to be converted into strings for template purposes.  That should be done below.

benchmark_runner/common/template_operations/template_operations.py
```diff
@@ -125,6 +128,7 @@ def generate_yamls_internal(self, scale: str = None, scale_node: str = None, red
             'standard_output_file': self.__standard_output_file,
             'standard_template_file': self.__standard_template_file,
             'scale': scale,
+            'scale_num': scale_num,
```
RobertKrawitz, line 131: This is where `scale` and `scale_num` or whatever you want to call it should be converted from numbers to strings.

benchmark_runner/common/template_operations/template_operations.py
```diff
@@ -181,7 +183,7 @@ def generate_yamls(self, scale: str = '', scale_nodes: list = [], redis: str = N
             if scale_nodes:
                 for bulk in bulks:
                     for num in bulk:
-                        p = Process(target=self.generate_yamls_internal, args=(str(num), scale_nodes[int(num/scale)], redis,))
+                        p = Process(target=self.generate_yamls_internal, args=(str(num), str(scale * len(scale_nodes)), scale_nodes[int(num/scale)], redis,))
```
RobertKrawitz, line 186: Again, just pass num and scale as integers, and let `generate_yamls_internal` convert them to strings.

RobertKrawitz (changes requested):

benchmark_runner/common/template_operations/template_operations.py
ebattat: Not see it as problem

benchmark_runner/common/template_operations/template_operations.py
RobertKrawitz: Casts should happen as close to the point of use as possible.  The caller shouldn't know the use to which the data is being put.  What if the callee needs to use it both ways?

ebattat: In this way, I cast it once and not several times

RobertKrawitz: Why does that matter?  This is not extremely performance-sensitive code where the cast value needs to be cached.

RobertKrawitz: If you need to cache something for performance reasons, that's one thing.  If not, it's a lot easier to read and maintain if it's passed around in its conceptual form and only cast at point of use.

benchmark_runner/common/template_operations/template_operations.py
```diff
@@ -94,12 +94,15 @@ def __build_template_data(self, previous_data: dict, template_root_data: dict):
                 **extra_data, **extra_kind_data, **extra_runtype_data, **extra_kind_runtype_data}
 
     @logger_time_stamp
-    def generate_yamls_internal(self, scale: str = None, scale_node: str = None, redis: str = None):
+    def __generate_yamls_internal(self, scale: str = None, scale_num: str = None, scale_node: str = None, redis: str = None):
```
RobertKrawitz, line 97: I still don't like casting it outside, but marking this function private guarantees that it is not an API that somebody else can use.

RobertKrawitz (approved):

PR #510, opened 2022-12-20
revert lso/odf fix

ebattat:
Fixes

- Revert LSO/ODF install changes for supporting OCP 4.12

Reason: odf 4.11 official version is not working with OCP 4.12, so revert it back to use patch market place version


RobertKrawitz (changes requested):
Please add a note to the description explaining why this needs to be reverted.

RobertKrawitz (approved):

PR #528, opened 2023-01-05
Set virtiofsd_threadpoolsize appropriately for vdbench

RobertKrawitz:
Set the threadpool size to the most I/O threads that will be used in a vdbench run.  If this improves performance, the same change will be made for perf-ci.

This is motivated by unexpected severe performance loss in vdbench runs on sandboxed containers (Kata) as compared to runc and OSV VMs.  Clusterbuster I/O runs do not show a similar loss of performance, and I hypothesize that this is the reason why.


ebattat (commented):
@RobertKrawitz, do you want me to test it on functional before merge ?

ebattat:
@RobertKrawitz, need to add to the description that the main reason for this change is drop in vdbench kata tests.

RobertKrawitz:
> @RobertKrawitz, do you want me to test it on functional before merge ?

Yes.

ebattat (approved):
/LGTM

PR #531, opened 2023-01-12
vdbench remove memory limit

ebattat:
Fix:

Remove memory limit of vdbench_kata due to memory leak issue when run kata in parallel in Perf-CI


RobertKrawitz (commented):
I think that this change should be specific to Kata.  I'm not going to hold up the PR over this, though.

ebattat:
@RobertKrawitz, please take a look now

RobertKrawitz (approved):

RobertKrawitz (approved):
At some point we may want to consider dispensing with memory limits for all tests for Kata, but right now this is the problem, so I'm OK with it.

PR #533, opened 2023-01-18
vm termination fix

ebattat:
Fix:

**The issue:** 
When terminating 600 VMs, some of the VMs stuck while terminating.
There is no way to terminate those VMs, the only way is to reinstall the cluster.

Known issue: https://github.com/kubernetes/kubernetes/issues/113426

**The solution:**
According to CNV developer we should update TerminationGracePeriodSeconds from 0 to 180

**The Explanation:**
By setting TerminationGracePeriodSeconds to 0 you've given your VM literally 0 seconds to complete shutdown gracefully. KubeVirt has no choice but to immediately send a kill command to your VM the instant the yaml is removed.

We suggest trying a number larger than 0 (180 should suffice) so the VM has time to complete I/O and shutdown properly before the timeout is reached and a kill signal is sent by the system.


gqlo (approved):
LGTM

PR #534, opened 2023-01-18
Bootstorm:add vm stop

ebattat:
Fix:
Add vm stop before terminate it.

This fix should solve 600 VMs termination stuck issue


RobertKrawitz (approved):
Question answered offline.

PR #535, opened 2023-01-25
disable functional environment

ebattat:
Fixes:

1. Disable Functional environment
2. Change the Perf CI to run on Tuesday and Friday, (Friday - running OpenShift Assisted installer)

### DIFF
--- a/.github/workflows/Nightly_Func_Env_CI.yml
+++ b/.github/workflows/Nightly_Func_Env_CI.yml
@@ -4,13 +4,13 @@ name: Deploy Func Env Nightly CI
 
 on:
   # Run on Nightly
-  schedule:
-    - cron: '0 4 * * 1-6' # run Monday-Saturday at 4 AM UTC/ 0 AM EDT
-  # Run on Sunday after assisted installer and operator completed
-  workflow_run:
-    workflows: ["Deploy Operator FUNC Env Weekly CI"]
-    types:
-      - completed
+#  schedule:
+#    - cron: '0 4 * * 1-6' # run Monday-Saturday at 4 AM UTC/ 0 AM EDT
+#  # Run on Sunday after assisted installer and operator completed
+#  workflow_run:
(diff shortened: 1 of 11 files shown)


.github/workflows/Nightly_Func_Env_CI.yml
```diff
@@ -4,13 +4,13 @@ name: Deploy Func Env Nightly CI
...
-    workflows: ["Deploy Operator FUNC Env Weekly CI"]
-    types:
-      - completed
+#  schedule:
+#    - cron: '0 4 * * 1-6' # run Monday-Saturday at 4 AM UTC/ 0 AM EDT
+#  # Run on Sunday after assisted installer and operator completed
+#  workflow_run:
+#    workflows: ["Deploy Operator FUNC Env Weekly CI"]
+#    types:
+#      - completed
```
RobertKrawitz, line 13: Should these (and all of the other functional environment actions) go away altogether?

RobertKrawitz (commented):
Only question is whether the func actions should be removed altogether, otherwise looks fine.

.github/workflows/Nightly_Func_Env_CI.yml
ebattat: Yes, we disable the cron job and calling by other workflow

RobertKrawitz: My question is whether we should remove these files.

ebattat: We can remove it but I dont see any problem to keep it as is

RobertKrawitz: Just to clean up unused code.  If we think we may want to bring it back, OK, but I'd like to keep everything tight.  I'm going to approve it; you decide whether you want to do that or not.

RobertKrawitz (approved):

PR #537, opened 2023-01-30
Remove memory limit pod

RobertKrawitz:
- Set memory limits to be identical for all runtimes (runc pod, VM, Kata pod).
- Add workaround for Kata not computing the memory limit correctly and thereby causing Kata pods doing heavy VM to crash.
- Include the failing template file if a template operation fails.
- Add a top level Makefile to rebuild and test the golden files to simplify development.

Verified with golden file check that the only changes are:
- vdbench runc and Kata pods now have a memory limit of 4Gi rather than 64Gi
- release, test_ci, and func_ci Kata pods now have a memory limit of 4Gi (previously had no memory limit), matching the equivalent runc pods


RobertKrawitz:
@ebattat @bbenshab Ready to review.
@jhopper None of the golden files for VMs have changed.  The VM templates did, in order to properly condition the resource requests/limits.  Note that the vdbench VM template previously set the memory limit to the value of the request, not that of the limit.

benchmark_runner/common/template_operations/templates/hammerdb/internal_data/hammerdb_vm_template.yaml
```diff
@@ -100,10 +100,14 @@ spec:
         cores: {{ cores }}
         threads: 1
         image: quay.io/cloud-bulldozer/{{ db_image }}
+        {%- if limits_memory %}
```
ebattat, line 103: @RobertKrawitz, I dont think we need those conditions because always we are using the limit, its also relate all the conditions below

jeniferh (approved):
LGTM
Makefile is a nice improvement, good catch on the mem limit getting set to req! 4Gi sounds good to me for vdbench

benchmark_runner/common/template_operations/templates/hammerdb/internal_data/hammerdb_vm_template.yaml
RobertKrawitz: Having these conditions gives us flexibility later if we want to remove them.

ebattat: For now we are going to keep the limits for all so I dont think we need it. it just makes the templates less readable, do you agree with it ?

RobertKrawitz: Hard to argue that it makes it harder to read.  Updated, but I'd like not to squash this out in case we want to do it later.

RobertKrawitz:
@ebattat before merging, we should apply the patch to the cluster to ensure that it does the right thing (and also because vdbench will fail on kata without this applied).  I've tested the patch manually, but that isn't exactly the same as testing it in action.

ebattat (approved):
/LGTM

PR #544, opened 2023-02-20
Workaround for actions/runner PR #1994 resulting in network error (connection reset by peer)

ebattat:
This PR is due to error we starting to get when using podman from GitHub actions.
This error is shown after running command that run more than 300 seconds.

Working properly till CI [02/01](https://github.com/redhat-performance/benchmark-runner/actions/runs/4021447216/jobs/6995719159)

First appeared at Installer CI [02/03](https://github.com/redhat-performance/benchmark-runner/actions/runs/3983220684/jobs/6828350542)

**Workaround for GitHub actions till fix according to [PR](https://github.com/actions/runner/issues/1994) **
`sudo -E XDG_RUNTIME_DIR= podman `

PR #545, opened 2023-02-20
Socket exception: Connection reset by peer (104)

ebattat:
This PR is due to error we starting to get when using podman from GitHub actions.
This error is shown after running command that run more than 300 seconds.

Working properly till CI [02/01](https://github.com/redhat-performance/benchmark-runner/actions/runs/4021447216/jobs/6995719159)

First appeared at Installer CI [02/03](https://github.com/redhat-performance/benchmark-runner/actions/runs/3983220684/jobs/6828350542)

Try to use docker after trying **Workaround for GitHub actions till fix according to https://github.com/actions/runner/issues/1994 **
`sudo -E XDG_RUNTIME_DIR= podman`
Ths problem in this [fix](https://github.com/redhat-performance/benchmark-runner/pull/544) is that ansible did not start, trying to run using docker


RobertKrawitz (commented):
Note in the comment and the PR that the workaround does not work, and that falling back to docker is the only method that works.  Also, I want you to update the actions issue to note that this workaround fails because Ansible will not run.

RobertKrawitz (approved):
Please update actions/runner/issues/1994 to note that the workaround is does not work for us as Ansible will not start.

PR #552, opened 2023-02-22
fix bootstorm dirty mem spike

ebattat:
Fix:
There is sometime a degradation in bootstorm vm login time due to  dirty memory spike.
Clean cache 3 times with 30 sec sleep between should solve the issue.

### DIFF
--- a/benchmark_runner/workloads/workloads_operations.py
+++ b/benchmark_runner/workloads/workloads_operations.py
@@ -23,6 +23,8 @@ from benchmark_runner.common.clouds.IBM.ibm_operations import IBMOperations
 class WorkloadsOperations:
     oc = None
     MILLISECONDS = 1000
+    REPEAT_TIMES = 3
+    SLEEP_TIME = 3
     """
     This class run workloads
     """
@@ -447,6 +449,9 @@ class WorkloadsOperations:
         """
         This method clear nodes cache
         """
+        for i in range(self.REPEAT_TIMES-1):
+            self._oc.clear_node_caches()
+            time.sleep(self.SLEEP_TIME)
         self._oc.clear_node_caches()
 
     def initialize_workload(self):


benchmark_runner/workloads/workloads_operations.py
```diff
@@ -447,7 +448,9 @@ def clear_nodes_cache(self):
         """
         This method clear nodes cache
         """
-        self._oc.clear_node_caches()
+        for i in range(self.REPEAT_TIMES):
+            self._oc.clear_node_caches()
+            time.sleep(self._bulk_sleep_time)
```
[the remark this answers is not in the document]
ebattat: Its 30 
I will create a constant for it

benchmark_runner/workloads/workloads_operations.py
```diff
@@ -447,7 +449,9 @@ def clear_nodes_cache(self):
         """
         This method clear nodes cache
         """
-        self._oc.clear_node_caches()
+        for i in range(self.REPEAT_TIMES):
+            self._oc.clear_node_caches()
+            time.sleep(self.SLEEP_TIME)
```
RobertKrawitz, line 454: 1. This should only sleep after the first N-1 repeat times.
2. 30 seconds between drop caches is excessive (90 seconds, or 60 when you fix (1)).  3 seconds should be fine.

RobertKrawitz (changes requested):

benchmark_runner/workloads/workloads_operations.py
ebattat: 1. I dont see any problem to do it after each cleaning, if it will not solve the issue we are going to revert it back.
2. I will change it to 3 sec

RobertKrawitz, line 454: Don't sleep after the final iteration.

RobertKrawitz (changes requested):

RobertKrawitz (approved):

PR #571, opened 2023-03-13
split namespace from workload for bootstorm and vdbench VMs

ebattat:
Fixes:

Vdbench vm: split namespace from workload
bootstorm vm:  split namespace from workload


RobertKrawitz (commented):
Why are there golden files for only some of the workload combinations?

ebattat:
@RobertKrawitz, I have changed the template only for vdbench vm/bootstorm vm.

RobertKrawitz (approved):

PR #583, opened 2023-04-24
Fix VM results parsing

ebattat:
Fix:

When start stamp didnt appear in the logs that causing parsing issue, adding another start stamp when podman call start.
This issue is happened when running vdbench VM is parallel.

### DIFF
--- a/benchmark_runner/common/oc/oc.py
+++ b/benchmark_runner/common/oc/oc.py
@@ -984,15 +984,15 @@ class OC(SSH):
             output_filename = os.path.join(self._run_artifacts, vm_name)
         results_list = []
         with open(output_filename) as infile:
-            copy = False
+            # VM logs results start
+            start = False
             for line in infile:
-                if start_stamp in line:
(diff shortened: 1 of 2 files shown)


ebattat:
@RobertKrawitz, the main reason for this issue is that start stamp is missing when Vdbench VM is loading very fast in multi-thread.
So we dont have a start stamp of capture logs and that affect the logs parser and that cause to empty results.
That the reason I added 'podman' in case of missing start stamp.

ebattat:
You can find below the logs w/ and w/o start stamp:

[vdbench_vm_without_start_stamp.log](https://github.com/redhat-performance/benchmark-runner/files/11313498/vdbench_vm_without_start_stamp.log)
[vdbench_vm_with_start_stamp.log](https://github.com/redhat-performance/benchmark-runner/files/11313499/vdbench_vm_with_start_stamp.log)

RobertKrawitz (approved):

PR #584, opened 2023-04-30
add describe pod vm and event logs

ebattat:
Fixes:

1. Add describe log for every running pod/vm (create also when pod/vm didnt create)
2. Add events log per run

### DIFF
--- a/benchmark_runner/common/oc/oc.py
+++ b/benchmark_runner/common/oc/oc.py
@@ -280,7 +280,7 @@ class OC(SSH):
     def create_async(self, yaml: str, is_check: bool = False):
         """
         This method create yaml in async
-        :param yaml:
+        @param yaml:
         @param is_check:
         :return:
         """
@@ -383,6 +383,15 @@ class OC(SSH):
             prom_token = self.run(f"{self.__cli} sa new-token -n openshift-monitoring prometheus-k8s 2>/dev/null")
         return prom_token
 
+    def collect_events(self):
+        """
+        This method collect event logs
+        :return: output_filename
+        """
+        output_filename = os.path.join(self._run_artifacts, f'events.logs')
+        self.run(f"{self.__cli} get events -A > '{output_filename}' ")
+        return output_filename
+
     @logger_time_stamp
     def login(self):
         """
@@ -425,12 +434,23 @@ class OC(SSH):
         """
         output_filename = os.path.join(self._run_artifacts, pod_name)
         if database:
-            self.run(f"{self.__cli} logs -n '{database}-db' {pod_name} > {output_filename} ")
+            self.run(f"{self.__cli} logs -n '{database}-db' {pod_name} > '{output_filename}' ")
         # manager logs of benchmark-controller-manager
         elif 'benchmark-controller-manager' in pod_name:
-            self.run(f"{self.__cli} logs -n {environment_variables.environment_variables_dict['namespace']} {pod_name} manager > {output_filename} ")
+            self.run(f"{self.__cli} logs -n {environment_variables.environment_variables_dict['namespace']} {pod_name} manager > '{output_filename}' ")
         else:
-            self.run(f"{self.__cli} logs -n {environment_variables.environment_variables_dict['namespace']} {pod_name} > {output_filename} ")
+            self.run(f"{self.__cli} logs -n {environment_variables.environment_variables_dict['namespace']} {pod_name} > '{output_filename}' ")
+        return output_filename
+
+    def describe_pod(self, pod_name: str, namespace: str = ''):
+        """
+        This method describe pod into log
+        :param pod_name: pod name with uuid
+        :param namespace: namespace
+        :return: output_filename
+        """
+        output_filename = os.path.join(self._run_artifacts, f'describe-{pod_name}')
+        self.run(f"{self.__cli} describe pod -n {namespace} {pod_name} > '{output_filename}' ")
         return output_filename
 
     @logger_time_stamp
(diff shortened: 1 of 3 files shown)


benchmark_runner/common/oc/oc.py
```diff
@@ -280,7 +280,7 @@ def apply_security_privileged(self, namespace: str = environment_variables.envir
     def create_async(self, yaml: str, is_check: bool = False):
         """
         This method create yaml in async
-        :param yaml:
+        @param yaml:
```
RobertKrawitz, line 283: What is the purpose of this?

benchmark_runner/common/oc/oc.py
```diff
@@ -383,6 +383,15 @@ def get_prom_token(self):
             prom_token = self.run(f"{self.__cli} sa new-token -n openshift-monitoring prometheus-k8s 2>/dev/null")
         return prom_token
 
+    def collect_events(self):
+        """
+        This method collect event logs
+        :return: output_filename
+        """
+        output_filename = os.path.join(self._run_artifacts, f'events.logs')
+        self.run(f"{self.__cli} get events -A > {output_filename} ")
```
RobertKrawitz, line 392: Quote and trailing space character:

```
self.run(f"'{self.__cli}' get events -A > '{output_filename}'")
```

benchmark_runner/common/oc/oc.py
```diff
@@ -433,6 +442,17 @@ def save_pod_log(self, pod_name: str, database: str = ''):
...
 
+    def describe_pod(self, pod_name: str, namespace: str = ''):
+        """
+        This method describe pod into log
+        :param pod_name: pod name with uuid
+        :param namespace: namespace
+        :return: output_filename
+        """
+        output_filename = os.path.join(self._run_artifacts, f'describe-{pod_name}')
+        self.run(f"{self.__cli} describe pod -n {namespace} {pod_name} > {output_filename} ")
```
RobertKrawitz, line 453: Quoting and trailing whitespace:

```
self.run(f"'{self.__cli}' describe pod -n '{namespace}' '{pod_name}' > '{output_filename}'")
```

benchmark_runner/common/oc/oc.py
```diff
@@ -692,6 +714,17 @@ def wait_for_pod_completed(self, label: str, workload: str = '', label_uuid: boo
...
 
+    def describe_vmi(self, vm_name: str, namespace: str):
+        """
+        This method describe vmi into log
+        :param vm_name: vm name with uuid
+        :param namespace: namespace
+        :return: output_filename
+        """
+        output_filename = os.path.join(self._run_artifacts, f'describe-{vm_name}')
+        self.run(f"{self.__cli} describe vmi -n {namespace} {vm_name} > {output_filename} ")
```
RobertKrawitz, line 725: Same quote and trailing whitespace as above.

RobertKrawitz (changes requested):
Some things need to be quoted.

benchmark_runner/common/oc/oc.py
ebattat: I have added it only to '{output_filename}', for 'oc' it will cause syntax's issue and also we pass this parameter.

benchmark_runner/common/oc/oc.py
ebattat: I have added it only to '{output_filename}', for 'oc' it will cause syntax's issue and also we pass this parameter.

benchmark_runner/common/oc/oc.py
ebattat: I have added it only to '{output_filename}', for 'oc' it will cause syntax's issue and also we pass this value.

benchmark_runner/common/oc/oc.py
ebattat: its syntax's mistake, parameter in comment should start with @ and not :

RobertKrawitz: OK, I guess this will eventually get fixed elsewhere.

benchmark_runner/common/oc/oc.py
RobertKrawitz: Have you tested that?

benchmark_runner/common/oc/oc.py
RobertKrawitz: See above.

benchmark_runner/common/oc/oc.py
ebattat: I dont see any reason to put 1 value parameter in quotes

ebattat: You right, for {output_filename} I have already added in all the places, but {self.__cli} is 1 parameter and not a file name.

RobertKrawitz: It could be, at least in principle.  Suppose self.__cli were `/usr/Here I Am/bin/oc`?

ebattat: The default value comes with [1 parameter](https://github.com/redhat-performance/benchmark-runner/blob/7f1c653d63006f234bc9a76d44af5ead3ff266da/benchmark_runner/main/environment_variables.py#L276-L282) no one should change it.

RobertKrawitz:
```
 self._environment_variables_dict['cli'] = EnvironmentVariables.get_env('CLI', 'oc')
```

What if someone sets CLI to something that has a space in it?

RobertKrawitz: Then why does it accept an environment variable at all?

ebattat: We can change it in case of major change, but its something that under our control

RobertKrawitz: So it's possible for the user to change.  There's no harm in programming defensively.

RobertKrawitz: (Strictly speaking, this isn't perfect either -- it's possible, although very rare, for a pathname to contain a quote character.  The really safe thing to do is to actually pass the arguments as a list, rather than as a string, but that's a separate issue.)

RobertKrawitz (approved):

PR #594, opened 2023-05-28
Fix Kata create issue

ebattat:
Fix:
This issue is started when upgrading 4.13.0-ec (based Rhel 8.6) to 4.13.0-rc (based Rhel 9.2).
The root cause is that the annotations: "--thread-pool-size=16"  is not working properly on Rhel9. 

The workaround is to update  "--thread-pool-size=16" in every node and not through annotations.

### DIFF
--- a/benchmark_runner/common/oc/oc.py
+++ b/benchmark_runner/common/oc/oc.py
@@ -142,6 +142,21 @@ class OC(SSH):
         """
         return self._get_kata_default_channel_field('currentCSVDesc.annotations."operatorframework.io/suggested-namespace"')
 
+    def set_kata_threads_pool(self, thread_pool_size: str):
+        """
+        This methods sets kata thread-pool-size in every worker node
+        @param thread_pool_size:
+        @return:
+        """
+        self.run(fr"""{self.__cli} get nodes -l node-role.kubernetes.io/worker= -o jsonpath="{{range .items[*]}}{{.metadata.name}}{{'\\n'}}{{end}}" |  xargs -I{{}} oc debug node/{{}} -- chroot /host sh -c "mkdir -p /etc/kata-containers; cp /usr/share/kata-containers/defaults/configuration.toml /etc/kata-containers/; sed -i 's/thread-pool-size=1/thread-pool-size={thread_pool_size}/' /etc/kata-containers/configuration.toml" """)
+
+    def delete_kata_threads_pool(self):
+        """
+        This methods deletes kata thread-pool-size from every worker node
+        @return:
+        """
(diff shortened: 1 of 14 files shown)


benchmark_runner/common/oc/oc.py
```diff
@@ -142,6 +142,21 @@ def _get_kata_namespace(self):
...
+        @return:
+        """
+        self.run(fr"""{self.__cli} get nodes -l node-role.kubernetes.io/worker= -o jsonpath="{{range .items[*]}}{{.metadata.name}}{{'\\n'}}{{end}}" |  xargs -I{{}} oc debug node/{{}} -- chroot /host sh -c "mkdir -p /etc/kata-containers; cp /usr/share/kata-containers/defaults/configuration.toml /etc/kata-containers/; sed -i 's/thread-pool-size=1/thread-pool-size={thread_pool_size}/' /etc/kata-containers/configuration.toml" """)
+
+    def delete_kata_threads_pool(self):
+        """
+        This methods deletes kata thread-pool-size from every worker node
+        @return:
+        """
+        self.run(fr"""{self.__cli} get nodes -l node-role.kubernetes.io/worker= -o jsonpath="{{range .items[*]}}{{.metadata.name}}{{'\\n'}}{{end}}" |  xargs -I{{}} oc debug node/{{}} -- chroot /host sh -c "rm -rf /etc/kata-containers" """)
```
RobertKrawitz, line 158: Only remove configuration.toml.  There may be other files in here that we should not touch.

benchmark_runner/common/oc/oc.py
```diff
@@ -142,6 +142,21 @@ def _get_kata_namespace(self):
...
+        @return:
+        """
+        self.run(fr"""{self.__cli} get nodes -l node-role.kubernetes.io/worker= -o jsonpath="{{range .items[*]}}{{.metadata.name}}{{'\\n'}}{{end}}" |  xargs -I{{}} oc debug node/{{}} -- chroot /host sh -c "mkdir -p /etc/kata-containers; cp /usr/share/kata-containers/defaults/configuration.toml /etc/kata-containers/; sed -i 's/thread-pool-size=1/thread-pool-size={thread_pool_size}/' /etc/kata-containers/configuration.toml" """)
+
+    def delete_kata_threads_pool(self):
+        """
+        This methods deletes kata thread-pool-size from every worker node
+        @return:
+        """
+        self.run(fr"""{self.__cli} get nodes -l node-role.kubernetes.io/worker= -o jsonpath="{{range .items[*]}}{{.metadata.name}}{{'\\n'}}{{end}}" |  xargs -I{{}} oc debug node/{{}} -- chroot /host sh -c "rm -rf /etc/kata-containers/configuration.toml" """)
```
RobertKrawitz, line 158: rm -f, not rm -rf in this case.

RobertKrawitz (changes requested):
Just one change.

RobertKrawitz (approved):

PR #597, opened 2023-06-04
Update jupyterlab templates

ebattat:
Fixes:

Update import hierarchy change from logs_analyzer => analyze_prometheus_logs


RobertKrawitz (commented):
Could you give this a better title?  The change is obvious, but I don't know why this is needed.

PR #599, opened 2023-06-07
fix lso issue

ebattat:
Fixes:

Allocate available pv for LSO workloads.
Currently we skipping available pv and that the reason that LSO workload is [failed](https://github.com/redhat-performance/benchmark-runner/actions/runs/5184460659/jobs/9371041195).


RobertKrawitz (changes requested):

ebattat:
We need to delete mount path when pv deleted

PR #601, opened 2023-06-08
remove LSO path

ebattat:
Fixed:

Removing LSO mount path at the beginning and ending of each LSO workload

** Currently Local Storage Operator didn't remove the mount path


RobertKrawitz (approved):
Is the LSO operator not removing the directory the intended behavior (in which case this is a bug fix) or is it a workaround for a bug (in which case it should be listed as such)?

I'm approving it as the code looks fine, but please update the text accordingly.

PR #602, opened 2023-06-08
add log type extension

ebattat:
Fix:

Adding extension for pod log.
For example if pod log is a csv file, update file type extension to CSV.
That is useful for log analyzing using JupyterLab.


benchmark_runner/common/oc/oc.py
```diff
@@ -417,10 +417,10 @@ def get_prom_token(self):
 
     def collect_events(self):
         """
-        This method collect event logs
+        This method collect event log
```
RobertKrawitz, line 420: collects

RobertKrawitz (approved):
One trivial typo in a comment.  I'm approving; please fix if possible before merging.

benchmark_runner/common/oc/oc.py

PR #603, opened 2023-06-08
fix Elasticsearch get delete id test

ebattat:
Fix:

Check if there is more than one document in Elasticsearch, that should solve the issue in case of previous data was left in Elasticsearch.

### DIFF
--- a/tests/integration/benchmark_runner/common/elasticsearch/test_elasticsearch_operations.py
+++ b/tests/integration/benchmark_runner/common/elasticsearch/test_elasticsearch_operations.py
@@ -50,12 +50,14 @@ def test_get_delete_ids_elasticsearch_last_day():
                                       es_port=test_environment_variable.get('elasticsearch_port', ''),
                                       es_user=test_environment_variable.get('elasticsearch_user', ''),
                                       es_password=test_environment_variable.get('elasticsearch_password', ''))
+    # upload data
     uuid = upload_data(elastic=elastic)
     elastic.verify_elasticsearch_data_uploaded(index=TEST_INDEX_NAME, uuid=uuid)
+    # verify data was uploaded to Elasticsearch
     current_timestamp = datetime.datetime.now()
     day_before_timestamp = current_timestamp - timedelta(days=1)
-    # verify data was uploaded to elastic
-    assert len(elastic.get_index_ids_between_dates(index=TEST_INDEX_NAME, start_datetime=day_before_timestamp, end_datetime=current_timestamp)) == 1
+    assert len(elastic.get_index_ids_between_dates(index=TEST_INDEX_NAME, start_datetime=day_before_timestamp, end_datetime=current_timestamp)) >= 1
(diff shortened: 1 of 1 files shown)


tests/integration/benchmark_runner/common/elasticsearch/test_elasticsearch_operations.py
```diff
@@ -50,12 +50,18 @@ def test_get_delete_ids_elasticsearch_last_day():
                                       es_port=test_environment_variable.get('elasticsearch_port', ''),
                                       es_user=test_environment_variable.get('elasticsearch_user', ''),
                                       es_password=test_environment_variable.get('elasticsearch_password', ''))
-    uuid = upload_data(elastic=elastic)
-    elastic.verify_elasticsearch_data_uploaded(index=TEST_INDEX_NAME, uuid=uuid)
     current_timestamp = datetime.datetime.now()
     day_before_timestamp = current_timestamp - timedelta(days=1)
+    # cleanup elastic
+    if len(elastic.get_index_ids_between_dates(index=TEST_INDEX_NAME, start_datetime=day_before_timestamp, end_datetime=current_timestamp)) >= 1:
+        elastic.delete_index_ids_between_dates(index=TEST_INDEX_NAME, start_datetime=day_before_timestamp, end_datetime=current_timestamp)
```
RobertKrawitz, line 57: I'm not sure I understand it.  It appears that you're looking for an index less than 24 hours old, and deleting it?  What happens if, say, multiple runs are done on one day, or a run ends less than 24 hours after the previous one did?

ebattat: The current issue is that if test failed the data keep remained in ES, so I do a cleanup before each test that ensure that each test will start with clean index in last 24 houes.

RobertKrawitz: What if the previous test succeded?  Or failed, but more than 24 hours ago?

Perhaps you should inject a token into ES at the very end of the run indicating that it succeeded; if it isn't there, then you know it failed.

ebattat: I have already verified that data upload into Elasticsearch properly in this [test](https://github.com/redhat-performance/benchmark-runner/blob/a37e24ebd63db69806fa3b0d3104ca0aeee6455a/tests/integration/benchmark_runner/common/elasticsearch/test_elasticsearch_operations.py#L35)

This test verified that data can be fetched or deleted in last 24 hours. 
I fixed the test and check if  >=1 document in Elasticsearch, that should solve the issue if there is a previous data.

RobertKrawitz: Oh I see, this is a test, not the real code.  My bad.  Approving.

PR #606, opened 2023-06-15
Work around virtiofsd threadpoolsize not working in 4.13

RobertKrawitz:
--threadpoolsize causes problems in OCP 4.13 because the Rust virtiofsd rejects duplicate command line arguments; the kata launcher always prepends --threadpoolsize=1, so we need to not add an extra argument of our own.  This will impact parallel I/O performance.

### DIFF
--- a/benchmark_runner/clusterbuster/clusterbuster_workloads.py
+++ b/benchmark_runner/clusterbuster/clusterbuster_workloads.py
@@ -94,10 +94,13 @@ class ClusterBusterWorkloads(WorkloadsOperations):
         :return:
         """
         # release pass uuid and workload
+        # WORKAROUND 20230615: --threadpoolsize causes failures in 4.13 due to Rust
+        # virtiofsd rejecting superseding arguments on the commandline, so we need
+        # to ensure that we suppress that for now.
         if self._run_type == 'release':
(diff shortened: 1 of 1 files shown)


benchmark_runner/clusterbuster/clusterbuster_workloads.py
```diff
@@ -94,10 +94,13 @@ def run_workload(self):
...
         else:
-            self.__ssh.run(cmd=f'{self.__clusterbuster_path} --run_type={self._run_type} --client-pin-node={self._pin_node1} --server-pin-node={self._pin_node2} --sync-pin-node={self._pin_node2} --force-cleanup-i-know-this-is-dangerous=600 --basename={self.__namespace} --artifactdir={self.__artifactdir} --analyze={self.__result_report} > {self.__clusterbuster_log} 2>&1')
+            self.__ssh.run(cmd=f'{self.__clusterbuster_path} --run_type={self._run_type} --client-pin-node={self._pin_node1} --server-pin-node={self._pin_node2} --sync-pin-node={self._pin_node2} --force-cleanup-i-know-this-is-dangerous=600 --basename={self.__namespace} --artifactdir={self.__artifactdir} --analyze={self.__result_report} > {self.__clusterbuster_log} --virtiofsd-threadpoolsize=0 2>&1')
         # Check that Result file exist and not empty
```
ebattat, line 104: --virtiofsd-threadpoolsize=0 , do we want to send hard coded 0 or to pass it by parameter ?

RobertKrawitz: --virtiofsd-threadpoolsize=0 is a special case.  We normally want it to be set automatically; in this case, we want to override that setting to not set it at all.

ebattat (approved):
/LGTM

PR #607, opened 2023-06-16
Fix the check for kata installation complete

RobertKrawitz:
Also fix the check for completed nodes; the field names were wrong.  It succeeded because both the total and completed node counts were always empty.


ebattat (approved):
/lgtm

RobertKrawitz:
Verified on my small cluster.

PR #612, opened 2023-06-29
Fix GitHub Actions PR vulnerability

ebattat:
Fix:

Adding ok-to-test label for running GitHub actions should prevent PR vulnerability

### DIFF
--- a/.github/workflows/Perf_Env_PR_Test_CI.yml
+++ b/.github/workflows/Perf_Env_PR_Test_CI.yml
@@ -6,6 +6,7 @@ name: Perf Env PR Test CI
 
 on:
   pull_request_target:
+    types: [labeled, opened]
     branches: [ main ]
   workflow_dispatch:
 
@@ -22,6 +23,8 @@ jobs:
     strategy:
       matrix:
         python-version: [ '3.7', '3.8', '3.9', '3.10' ]
+    # minimize potential vulnerabilities
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
     steps:
       - uses: actions/checkout@v3
         with:
--- a/tests_requirements.txt
+++ b/tests_requirements.txt
@@ -7,4 +7,4 @@ moto==2.2.16
 pandas
 pytest
 PyYAML==6.0
-typeguard
+typeguard==2.12.1


tests_requirements.txt
```diff
@@ -7,4 +7,4 @@ moto==2.2.16
 pandas
 pytest
 PyYAML==6.0
-typeguard
+typeguard==2.12.1
```
RobertKrawitz, line 10: Get rid of this file (and requirements.txt) altogether if we aren't going to use them.

.github/workflows/Perf_Env_PR_Test_CI.yml
```diff
@@ -181,8 +185,10 @@ jobs:
      run: |
        python -m pip install --upgrade pip
        pip install flake8 pytest pytest-cov
-       if [[ -f requirements.txt ]]; then pip install -r requirements.txt; fi
-       if [[ -f tests_requirements.txt ]]; then pip install -r tests_requirements.txt; fi
+       # install requirements.txt 
+       pip install attrs==21.4.0 azure==4.0.0 boto3==1.26.1 botocore==1.29.1 cryptography==41.0.0 elasticsearch==7.16.1 elasticsearch_dsl==7.4.0 ipywidgets==8.0.6 jinja2==3.0.3 myst-parser==0.17.0 openshift-client==1.0.17 prometheus-api-client==0.5.1 pandas paramiko==3.0.0 PyGitHub==1.55 PyYAML==6.0 sphinx==4.5.0 sphinx_rtd_theme==1.0.0 tenacity==8.0.1 tqdm==4.65.0 typeguard==2.12.1 typing==3.7.4.3
+       # install tests_requirements.txt
+       pip install mock moto==2.2.16 pytest
```
RobertKrawitz, line 191: Specify the version for mock and pytest.  Also, this list is a lot shorter than tests_requirements.txt.

RobertKrawitz (changes requested):
Should get rid of requirements.txt and tests_requirements.txt, and add a note to the documentation about how to correctly add new requirements.

tests_requirements.txt
ebattat: test_requirement.txt is for testing propose and it should be isolate for requirement.txt, because we dont want to mix the test pkgs with project pkg.

RobertKrawitz: If we're going to hard code the packages and versions in the pip command line, we should not have either requirements.txt file, just the pip commands.

.github/workflows/Perf_Env_PR_Test_CI.yml
RobertKrawitz: OK, just be aware of the tradeoffs if mock or pandas changes something.

tests_requirements.txt
ebattat: We need to add it only to the PR yaml file to mitigate risk.
We are still use it regular when [merging to main](https://github.com/redhat-performance/benchmark-runner/blob/90483db157edbcee2efd82ccb36415cea3a33895/.github/workflows/Perf_Env_Build_Test_CI.yml#L35-L36)

.github/workflows/Perf_Env_PR_Test_CI.yml
sadsfae: How many of the pip installed packages could just be satisfied with `apt install` ?  This way you'd gain the benefit of natural evolution in package-sets and versions from `ubuntu-latest`.

ebattat: I dont see any risk to keep it using `pip install`, the major risk behind us.

sadsfae: > How many of the pip installed packages could just be satisfied with `apt install` ? This way you'd gain the benefit of natural evolution in package-sets and versions from `ubuntu-latest`.

I found replacements for almost all of your `requirements.txt` pip packages and other packages except a few.

== From packages search:  https://packages.ubuntu.com/ ==

`python -m pip install --upgrade pip`
`pip install flake8 pytest pytest-cov`

Can be replaced by

`apt install python3-pip` 
`apt install python3-flake8 python3-pytest python3-pytest-cov`

== From requirements ==
https://raw.githubusercontent.com/redhat-performance/benchmark-runner/main/requirements.txt

| Pip Package | Apt Package |
| ------------------- | ---------------------|
|attrs | python3-attr |
|azure | **no match** |
|boto3 | python3-boto3 |
|botocore | python3-botocore |
|cryptography | python3-cryptography |
|elasticsearch | python3-elasticsearch |
|elasticsearch_dsl | **no match** |
|ipywidgets | python3-ipywidgets |
|jinja2 | python3-jinja2 |
|myst-parser | python3-myst-parser |
|openshift-client | python3-openshift |
|prometheus-api-client | python3-prometheus-client? |
|pandas | python3-pandas |
|paramiko | python3-paramiko |
|PyGithub | **no match** |
|sphinx | sphinx-common |
|sphinx_rtd_theme | sphinx-rtd-theme-common |
|tenacity | python3-tenacity |
|tqdm | python3-tqdm |
|typeguard | python3-typeguard |
|typing | python3-typing-extensions |

You'd still need some pip packages in this case unless the ones without a match aren't needed.  I would say if you can't replace _all_ of your pip packages with system packages maybe it's not worth mixing them.

The risk with using pip at all is with having `setup.py` available for someone to modify in a pull request, it's part of your repo so it can be modified and run with the GHA access token against your repository.

**Why Pip and setup.py isn't Safe with pull_request_target**

The Github Actions `pull_request_target` trigger runs against _your_ repository, not their fork.  It has full write permissions because it uses the GHA access token on the Github side.  Nothing stops someone from modifying `setup.py` to execute pre-install scripts before pip runs in your GHA yaml since shell commands are available.

`pull_request_target` is a workflow trigger that was added later, different from `pull_request`

I think since you can't entirely switch from pip to system packages due to a few missing above that I didn't see apt packages for perhaps a workaround is:

1) Create a different branch that **doesn't** include `setup.py`
2) Have pull requests only able to checkout and PR from this branch
3) This branch gets merged up minus `setup.py`

Perhaps you don't even need `setup.py` to begin with if you already have explicit pip install commands.

ebattat: @sadsfae, the main idea is to run unittest and integration tests using the exact python package versions, otherwise all the tests we run in the PR are meaningless ad we can delete all this file. I think that the current fix should reduce the risk to inject python packages into requirments.txt/test_requirments.txt.

sadsfae:
I think this setting is also helpful anyway if getting PR from unknown person:

https://github.blog/changelog/2021-04-22-github-actions-maintainers-must-approve-first-time-contributor-workflow-runs/

.github/workflows/Perf_Env_PR_Test_CI.yml
AdnaneKhan: Don't mean to intrude into your conversation in resolving this; however, one great way to stop `pull_request_target` abuse is to add a check for an "ok to test" label before running any tests from fork PRs. You can do it like this:

First make it so the event triggers on PR opened as well as labeled
```
on:
  pull_request_target:
    types: [labeled, opened]
    branches: [main]
    
```

then you can add an if statement on the first job to check for the label. You could also allow the label placed by dependabot to automatically run  workflows for those PRs.

```
 if: ${{ contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
```

ebattat: @AdnaneKhan thanks for your elegant secured way to solve this issue.
I have tested it [locally](https://github.com/ebattat/Test/pull/13) and it works perfect including PR auto run when change the label to **ok-to-test**.
BTW, I have added the if condition only **once** in unittest job because other jobs: integration_test and e2e need it.
If that ok and secured I will do the same fix in [cloud-governance project](https://github.com/redhat-performance/cloud-governance).

sadsfae (approved):
This looks good and also appreciate @AdnaneKhan suggestions for further protections with `pull_request_target`

RobertKrawitz (approved):
Approved from my side, needs security signoff.

.github/workflows/Perf_Env_PR_Test_CI.yml
sectest-h1-adnanthekhan: Yeah, the label will protect you from a malicious workflow as long as you properly inspect the changes before adding the label.

There is a *slight* risk if someone were to create a fork, make legitimate changes, get it labeled, and then push more malicious changes. However, since your project doesn't take direct PRs from the public you wouldn't label the PR to begin with.

athiruma (approved):
lgtm

.github/workflows/Perf_Env_PR_Test_CI.yml
ebattat: @AdnaneKhan, thanks for this useful security recommendations.
Same [fix](https://github.com/redhat-performance/cloud-governance/pull/630) will add to cloud-governance repository

PR #617, opened 2023-07-13
reduce fedora VMs number

ebattat:
Fixed:
Currently we use 100GB from 128GB when running 100 fedora VMs.
That cause to not stable results in regressions.
The recommendation is to reduce the number of VMs to 80.

### DIFF
--- a/.github/workflows/Nightly_Perf_Env_CI.yml
+++ b/.github/workflows/Nightly_Perf_Env_CI.yml
@@ -175,7 +175,7 @@ jobs:
         WORKER_DISK_IDS: ${{ secrets.PERF_WORKER_DISK_IDS }}
         TIMEOUT: 8000
         SCALE: 2
-        BOOTSTROM_SCALE: 100
+        BOOTSTORM_SCALE: 80
         WINDOWS_SCALE: 40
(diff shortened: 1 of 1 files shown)


jeniferh:
LGTM. I hope this proves to be more stable!

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -175,7 +175,7 @@ jobs:
         WORKER_DISK_IDS: ${{ secrets.PERF_WORKER_DISK_IDS }}
         TIMEOUT: 8000
         SCALE: 2
-        BOOTSTROM_SCALE: 100
+        BOOTSTROM_SCALE: 80
```
RobertKrawitz, line 178: While you're at it, please fix `BOOTSTROM_SCALE` to `BOOTSTORM_SCALE`.

RobertKrawitz (changes requested):

.github/workflows/Nightly_Perf_Env_CI.yml
ebattat: Thx done

RobertKrawitz (approved):

PR #618, opened 2023-07-13
Remove open pr

ebattat:
Fixed:
Not need to run GitHub actions against PR when its opened.
It will run auto after changing the label to ok-to-test

### DIFF
--- a/.github/workflows/Nightly_Perf_Env_CI.yml
+++ b/.github/workflows/Nightly_Perf_Env_CI.yml
@@ -175,7 +175,7 @@ jobs:
         WORKER_DISK_IDS: ${{ secrets.PERF_WORKER_DISK_IDS }}
         TIMEOUT: 8000
         SCALE: 2
-        BOOTSTROM_SCALE: 100
+        BOOTSTORM_SCALE: 80
         WINDOWS_SCALE: 40
         THREADS_LIMIT: 20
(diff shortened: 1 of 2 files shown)


.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -175,7 +175,7 @@ jobs:
         WORKER_DISK_IDS: ${{ secrets.PERF_WORKER_DISK_IDS }}
         TIMEOUT: 8000
         SCALE: 2
-        BOOTSTROM_SCALE: 100
+        BOOTSTORM_SCALE: 80
```
RobertKrawitz, line 178: This should not be in this PR.

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -192,7 +192,7 @@ jobs:
             # bootstorm_vm_scale: no need redis for synchronization but need SCALE and THREADS_LIMIT
             if [[ '${{ matrix.workload }}' == 'bootstorm_vm_scale' ]]
             then
-              SCALE=$BOOTSTROM_SCALE
+              SCALE=$BOOTSTORM_SCALE
```
RobertKrawitz, line 195: This should not be in this PR.

RobertKrawitz (changes requested):

PR #619, opened 2023-07-13
remove opened pr

ebattat:
Fixed:
Not need to run GitHub actions against PR when its opened.
It will run auto after changing the label to ok-to-test


RobertKrawitz (approved):

athiruma (approved):
/LGTM

PR #621, opened 2023-07-17
fix large template file issue

ebattat:
Fix:

1. When displayed large Jupyterlab > 1MB templates got the following error:
```
File Save Error for presentation.ipynb
Invalid response: 413
```
Solved it according to the following [recommendation](https://github.com/nds-org/esiphub/issues/18)

Adding the following parameter to Jupyterlab section in nginx conf:
`client_max_body_size 64m;`


athiruma (approved):
/lgtm

PR #629, opened 2023-07-28
install kata custom version

ebattat:
Fix:
Add ability to install custom kata version instead of latest version. [Required when there is an issue in latest version]

### DIFF
--- a/.github/workflows/Weekly_Func_Env_Operator_CI.yml
+++ b/.github/workflows/Weekly_Func_Env_Operator_CI.yml
@@ -80,6 +80,8 @@ jobs:
           CNV_VERSION: "4.13"
           ODF_VERSION: "4.12"
           NUM_ODF_DISK : "4"
+          KATA_CHANNEL: "stable-1.4"
+          KATA_CSV: "sandboxed-containers-operator.v1.4.0"
           QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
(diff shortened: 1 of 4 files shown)


.github/workflows/Weekly_Func_Env_Operator_CI.yml
```diff
@@ -80,6 +80,8 @@ jobs:
           CNV_VERSION: "4.13"
           ODF_VERSION: "4.12"
           NUM_ODF_DISK : "4"
+          KATA_CHANNEL: "stable-1.4"
```
RobertKrawitz, line 83: Need a comment that this is a temporary workaround here and elsewhere.

RobertKrawitz (changes requested):
The PR text needs to note that this is a short-term workaround, and this needs to be reverted (at least the places that set it) as soon as the underlying problem is fixed.

.github/workflows/Weekly_Func_Env_Operator_CI.yml
ebattat: Its not temporary, once 1.4.1 issue fix I will remove just the value and I will leave all the rest as is

RobertKrawitz: Discussed offline and agree with the current wording.

RobertKrawitz (approved):

PR #635, opened 2023-08-02
add kata 141 patch

ebattat:
bug:
Adding kata operator 141 patch, till issue will be solved


RobertKrawitz:
We should make the CI robust against this kind of installation failure rather than using temporary workarounds.

PR #641, opened 2023-08-07
PerfCi Grafana Dashboard: add grafonnet support

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Adding grafonnet support:

1. Dockerfile: using **terraform** for generating dashboard.json from grafonnet (main.libsonnet)
2. Update Grafana dashboard using generated dashboard.json 
3. Updating automatically new product versions from ElasticSearch and update main.libsonnet
4. Update Build and Nightly workflows with update_grafana_dashboard job that update Grafana dashboard
5. Add test for fetching all dashboards from Grafana for testing connectivity.
6. jsonnet/jsonnetfile.lock.json: This file configure specific grafonnet version and not with latest, that will reduce grafonnet generator failures.

[ **Important**: Currently the implementation is against testing dashboard, once all working update PerfCi dashboard ]


benchmark_runner/common/grafana/update_grafana_last_value_mappings.py
```diff
@@ -0,0 +1,117 @@
...
+        # Pattern to check in each line
+        start = "        stateTimeline.valueMapping.RegexMap.withOptions("
+        end = "+ stateTimeline.valueMapping.RegexMap.withType('value')"
+
+        # Initialize a flag to indicate whether the lines are within the desired range
+        within_range = False
+        mapping_lines = []
+
+        # Read the content of the file and store lines between start and end patterns in a list
+        with open(self.main_libsonnet_path, 'r') as file:
```
RobertKrawitz, line 58: Ugh.  You're editing Python code this way?  If that's what you're doing, I suggest using a clearly named comment that won't be duplicated at the end of the line, something like
```
# START_92734987
```
or some other fixed cookie.

benchmark_runner/common/grafana/update_grafana_last_value_mappings.py
```diff
@@ -0,0 +1,117 @@
...
+                self.value_mappings[key] = {"index": int(max_index) + num, "text": value}
+
+    def update_main_libsonnet(self):
+        """
+        This method updated last versions in main.libsonnet
+        :return:
+        """
+        # Pattern to check in each line
+        start = "        stateTimeline.valueMapping.RegexMap.withOptions("
+        end = "+ stateTimeline.valueMapping.RegexMap.withType('value')"
```
RobertKrawitz, line 98: Again, find a more robust way to do this.

benchmark_runner/common/grafana/update_grafana_last_value_mappings.py
```diff
@@ -0,0 +1,117 @@
...
+        end_index = None
+        for i, line in enumerate(lines):
+            if start in line:
+                start_index = i
+            elif end in line:
+                end_index = i
+
+        # Replace the lines between start and end with the dictionary
+        if start_index is not None and end_index is not None:
+            lines[start_index + 1:end_index] = ['\t\t\t\t' + json.dumps(self.value_mappings, indent=None) + '\n \t\t) \n']
```
RobertKrawitz, line 113: Why do you need the tabs?  And if you do, you do not need a space after the newline immediately before the tab.

benchmark_runner/grafana/perf/Makefile
```diff
@@ -0,0 +1,6 @@
+.PHONY: apply
+apply:
+	@cd jsonnet/ && \
+		jb install && \
+		cd ../ && \
+		terraform apply -auto-approve
```
RobertKrawitz, line 6: Need a newline at the end.

benchmark_runner/grafana/perf/dashboard.tf
```diff
@@ -0,0 +1,20 @@
...
+  jsonnet_path = "${path.cwd}/jsonnet/vendor"
+}
+
+data "jsonnet_file" "dashboard" {
+  source = "${path.cwd}/jsonnet/main.libsonnet"
+}
+
+output "dashboard" {
+  value = data.jsonnet_file.dashboard.rendered
+}
```
RobertKrawitz, line 20: Add trailing newline.

benchmark_runner/grafana/perf/extract_dashboard.py
```diff
@@ -0,0 +1,10 @@
+import re
+import argparse
+
+parser = argparse.ArgumentParser(description="Extract text between <<EOT and EOT in input file.")
```
RobertKrawitz, line 4: Is this supposed to extract it even if not a line of its own (in the latter case) and at the end of the line (in the former)?

benchmark_runner/grafana/perf/jsonnet/g.libsonnet
```diff
@@ -0,0 +1 @@
+import 'github.com/grafana/grafonnet/gen/grafonnet-latest/main.libsonnet'
```
RobertKrawitz, line 1: Is there something special about g.libsonnet that you can't just have something else use main.gsonnet?

benchmark_runner/grafana/perf/jsonnet/jsonnetfile.json
```diff
@@ -0,0 +1,15 @@
...
+        "git": {
+          "remote": "https://github.com/grafana/grafonnet.git",
+          "subdir": "gen/grafonnet-latest"
+        }
+      },
+      "version": "main"
+    }
+  ],
+  "legacyImports": true
+}
```
RobertKrawitz, line 15: Add trailing newline

benchmark_runner/grafana/perf/jsonnet/main.libsonnet
```diff
@@ -0,0 +1,7045 @@
+local g = import 'g.libsonnet';
```
RobertKrawitz, line 1: Not reviewing this file.

benchmark_runner/grafana/update_grafana_dashboard.py
```diff
@@ -0,0 +1,24 @@
...
+grafana_operations = GrafanaOperations(grafana_url=environment_variables_dict.get('grafana_url', ''),
+                                       grafana_api_key=environment_variables_dict.get('grafana_api_key', ''),
+                                       grafana_json=environment_variables_dict.get('grafana_json', ''))
+# for debug: grafana_operations.fetch_all_dashboards()
+grafana_operations.read_dashboard_json()
+grafana_operations.increment_dashboard_version()
+grafana_operations.override_dashboard()
+
+# Error: 412
+# The 412 status code is used when a newer dashboard already exists (newer, its version is greater than the version that was sent). The same status code is also used if another dashboard exists with the same title.
```
RobertKrawitz, line 24: What does this comment refer to (this looks like the entire file)?

tests/integration/benchmark_runner/common/grafana/test_grafana_operations.py
```diff
@@ -0,0 +1,18 @@
...
+    @return:
+    """
+    if test_environment_variable.get('grafana_url'):
+        grafana_operations = GrafanaOperations(grafana_url=test_environment_variable.get('grafana_url', ''),
+                                               grafana_api_key=test_environment_variable.get('grafana_api_key', ''),
+                                               grafana_json=test_environment_variable.get('grafana_json', ''))
+        all_dashboards = grafana_operations.fetch_all_dashboards()
+        assert all_dashboards
+    else:
+        raise Exception('incorrect grafana url')
```
RobertKrawitz, line 18: Add trailing newline.

RobertKrawitz (commented):
You may want Thiru and/or Kelly to review this too.

RobertKrawitz (changes requested):
Needs a number of changes as specified.

tests/integration/benchmark_runner/common/grafana/test_grafana_operations.py
ebattat: Fixed

benchmark_runner/grafana/update_grafana_dashboard.py
ebattat: If We got this error, to explain the reason

benchmark_runner/grafana/perf/jsonnet/main.libsonnet
ebattat: Yes, its the grafonnet code

benchmark_runner/grafana/perf/jsonnet/jsonnetfile.json
ebattat: Fixed

benchmark_runner/grafana/perf/jsonnet/g.libsonnet
ebattat: All is according to [grafonnet](https://github.com/grafana/grafonnet/blob/main/examples/terraform/jsonnet/g.libsonnet)

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -217,11 +217,58 @@ jobs:
...
+          cp $GITHUB_WORKSPACE/benchmark_runner/grafana/perf/dashboard_backup.json $GITHUB_WORKSPACE/benchmark_runner/grafana/perf/dashboard.json
+
+      - name: Commit dashboard.json if any changes are detected
+        if: steps.check_changes.outputs.changes == 'true'
+        run: |
+          echo ${{ steps.check_changes.outputs.changes }}
+          git checkout main
+          git config --global user.email ${{ secrets.EMAIL }}
+          git config --global user.name  ${{ secrets.USER_NAME }}
+          git add $GITHUB_WORKSPACE/benchmark_runner/grafana/perf/dashboard.json
```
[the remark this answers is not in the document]
ebattat: Fixed

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -217,11 +217,58 @@ jobs:
...
+          GRAFANA_JSON: ${{ secrets.PERF_GRAFANA_JSON }}
+        run: |
+          python $GITHUB_WORKSPACE/benchmark_runner/grafana/update_grafana_dashboard.py
+          # revert dashboard.json after python version change
+          cp $GITHUB_WORKSPACE/benchmark_runner/grafana/perf/dashboard_backup.json $GITHUB_WORKSPACE/benchmark_runner/grafana/perf/dashboard.json
+
+      - name: Commit dashboard.json if any changes are detected
+        if: steps.check_changes.outputs.changes == 'true'
+        run: |
+          echo ${{ steps.check_changes.outputs.changes }}
```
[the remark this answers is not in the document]
ebattat: Yes, Removed

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -217,11 +217,58 @@ jobs:
...
+      - name: Update the Grafana dashboard if any changes are detected in dashboard.json
+        if: steps.check_changes.outputs.changes == 'true'
+        env:
+          GRAFANA_URL: ${{ secrets.PERF_GRAFANA_URL }}
+          GRAFANA_API_KEY: ${{ secrets.PERF_GRAFANA_API_KEY }}
+          GRAFANA_JSON: ${{ secrets.PERF_GRAFANA_JSON }}
+        run: |
+          python $GITHUB_WORKSPACE/benchmark_runner/grafana/update_grafana_dashboard.py
+          # revert dashboard.json after python version change
+          cp $GITHUB_WORKSPACE/benchmark_runner/grafana/perf/dashboard_backup.json $GITHUB_WORKSPACE/benchmark_runner/grafana/perf/dashboard.json
```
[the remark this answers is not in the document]
ebattat: Fixed

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -217,11 +217,58 @@ jobs:
...
+
+      - name: Generate grafana dashboard.json using grafonnet container
+        run: |
+          sudo docker run --rm --name run_grafonnet -v $GITHUB_WORKSPACE/benchmark_runner/grafana/perf:/app --privileged quay.io/ebattat/run_grafonnet:latest
+
+      - name: Check for changes in dashboard.json
+        id: check_changes
+        run: |
+          # backup dashboard.json
+          cp $GITHUB_WORKSPACE/benchmark_runner/grafana/perf/dashboard.json $GITHUB_WORKSPACE/benchmark_runner/grafana/perf/dashboard_backup.json
```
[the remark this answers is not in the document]
ebattat: Fixed

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -217,11 +217,58 @@ jobs:
...
+    name: update_grafana_dashboard
+    needs: [ initialize_nightly, workload ]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Generate grafana dashboard.json using grafonnet container
+        run: |
+          sudo docker run --rm --name run_grafonnet -v $GITHUB_WORKSPACE/benchmark_runner/grafana/perf:/app --privileged quay.io/ebattat/run_grafonnet:latest
```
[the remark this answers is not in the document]
ebattat: Fixed

benchmark_runner/grafana/update_grafana_dashboard.py
RobertKrawitz: It's just a comment at the very end of a file.  This comment should go with whatever can actually produce that error, or better yet, if the error is received, there should be an error message.

.github/workflows/Nightly_Perf_Env_CI.yml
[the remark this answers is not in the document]
ebattat: Update the reason of file backup in comment

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -217,11 +217,57 @@ jobs:
...
+
+      - name: Generate grafana dashboard.json using grafonnet container
+        run: |
+          sudo podman run --rm --name run_grafonnet -v "$GITHUB_WORKSPACE"/benchmark_runner/grafana/perf:/app --privileged quay.io/ebattat/run_grafonnet:latest
+
+      - name: Check for changes in dashboard.json
+        id: check_changes
+        run: |
+          # Backup dashboard.json before updating the Grafana dashboard to avoid changes in file indentation
+          cp -p "$GITHUB_WORKSPACE"/benchmark_runner/grafana/perf/dashboard.json "$GITHUB_WORKSPACE"/benchmark_runner/grafana/perf/dashboard_backup.json
```
RobertKrawitz, line 237: Can't you put this line in the update Grafana dashboard immediately before the `python "$GITHUB...` line, before the file could be modified?

benchmark_runner/grafana/update_grafana_dashboard.py
ebattat: You right, Move the error code message it to grafana_operation.py where it raised

.github/workflows/Nightly_Perf_Env_CI.yml
ebattat: Fixed it

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -217,11 +217,57 @@ jobs:
...
+        uses: actions/checkout@v3
+
+      - name: Generate grafana dashboard.json using grafonnet container
+        run: |
+          sudo podman run --rm --name run_grafonnet -v "$GITHUB_WORKSPACE"/benchmark_runner/grafana/perf:/app --privileged quay.io/ebattat/run_grafonnet:latest
+
+      - name: Check for changes in dashboard.json
+        id: check_changes
+        run: |
+          git diff --quiet "$GITHUB_WORKSPACE"/benchmark_runner/grafana/perf/dashboard.json || echo "changes=true" >> $GITHUB_OUTPUT
```
RobertKrawitz, line 236: "$GITHUB_OUTPUT" also needs to be quoted.

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -217,11 +217,58 @@ jobs:
...
+        run: |
+          echo ${{ steps.check_changes.outputs.changes }}
+          git checkout main
+          git config --global user.email ${{ secrets.EMAIL }}
+          git config --global user.name  ${{ secrets.USER_NAME }}
+          git add $GITHUB_WORKSPACE/benchmark_runner/grafana/perf/dashboard.json
+          git commit -m "Update grafana json file"
+
+          # Pull the latest changes from the remote main branch
+          git pull origin main
```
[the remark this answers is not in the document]
RobertKrawitz: Comment still applies.

.github/workflows/Perf_Env_Build_Test_CI.yml
```diff
@@ -261,9 +264,54 @@ jobs:
...
+        uses: actions/checkout@v3
+
+      - name: Generate grafana dashboard.json using grafonnet container
+        run: |
+          sudo podman run --rm --name run_grafonnet -v "$GITHUB_WORKSPACE"/benchmark_runner/grafana/perf:/app --privileged quay.io/ebattat/run_grafonnet:latest
+
+      - name: Check for changes in dashboard.json
+        id: check_changes
+        run: |
+          git diff --quiet "$GITHUB_WORKSPACE"/benchmark_runner/grafana/perf/dashboard.json || echo "changes=true" >> $GITHUB_OUTPUT
```
RobertKrawitz, line 282: "$GITHUB_OUTPUT"

benchmark_runner/common/grafana/grafana_operations.py
```diff
@@ -0,0 +1,102 @@
+
+import requests
+import json
+
+
+class GrafanaOperations:
+    """
+    This class responsible for Grafana operations
```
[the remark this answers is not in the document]
RobertKrawitz: Comment still applies.

benchmark_runner/common/grafana/grafana_operations.py
```diff
@@ -0,0 +1,102 @@
...
+        self.grafana_json = grafana_json
+        self.dashboard_data = {}
+
+    def fetch_all_dashboards(self):
+        """
+        This method fetches all dashboards
+        :return:
+        """
+        dashboard_list = []
+        headers = {
```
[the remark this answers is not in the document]
RobertKrawitz: Comment still applies.

benchmark_runner/grafana/perf/jsonnet/g.libsonnet
```diff
@@ -0,0 +1,2 @@
+// Cannot import it in main.libsonnet; it must be in a separate file, g.libsonnet for more details: https://github.com/grafana/grafonnet/tree/main/examples/terraform
```
RobertKrawitz, line 1: That's just a pointer to an example.  Note that attempting to import it inside main.libsonnet results in an error.

RobertKrawitz (changes requested):
Still have some comments to address.

benchmark_runner/common/grafana/update_grafana_last_value_mappings.py
ebattat: Good point, fix it by adding
```
        //START_VALUE_MAPPING_227
        //END_VALUE_MAPPING_227

```

benchmark_runner/common/grafana/update_grafana_last_value_mappings.py
ebattat: Fixed, see above

benchmark_runner/common/grafana/update_grafana_last_value_mappings.py
ebattat: need to indent the value mapping

benchmark_runner/grafana/perf/Makefile
ebattat: Fixed

benchmark_runner/grafana/perf/dashboard.tf
ebattat: Fixed

benchmark_runner/grafana/perf/extract_dashboard.py
ebattat: The terrafom will be failed if no line

benchmark_runner/grafana/perf/jsonnet/g.libsonnet
ebattat: Updated

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -314,3 +315,49 @@ jobs:
...
+          python "$GITHUB_WORKSPACE"/benchmark_runner/grafana/update_grafana_dashboard.py
+          # Revert dashboard.json after adding dashboard version by Python code
+          cp -p "$GITHUB_WORKSPACE"/benchmark_runner/grafana/perf/dashboard_backup.json "$GITHUB_WORKSPACE"/benchmark_runner/grafana/perf/dashboard.json
+
+      - name: Commit dashboard.json if any changes are detected
+        if: steps.check_changes.outputs.changes == 'true'
+        run: |
+          git checkout main
+          git config --global user.email ${{ secrets.EMAIL }}
+          git config --global user.name  ${{ secrets.USER_NAME }}
```
RobertKrawitz, line 355: Should be

```
git config --global user.email "${{ secrets.EMAIL }}"
```

benchmark_runner/common/grafana/grafana_operations.py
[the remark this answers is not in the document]
RobertKrawitz: Shouldn't the http request have a Content-Type: header since you expect JSON?

benchmark_runner/common/grafana/update_grafana_last_value_mappings.py
```diff
@@ -0,0 +1,117 @@
...
+
+        ids = self.elasticsearch.get_index_ids_between_dates(index='ci-status', start_datetime=start_datetime,
+                                                        end_datetime=current_datetime)
+
+        display_versions = ['ocp_version', 'cnv_version', 'kata_version', 'kata_rpm_version', 'odf_version']
+        for id in ids:
+            data = self.elasticsearch.get_elasticsearch_index_by_id(index='ci-status', id=id)
+            for version, value in data['_source'].items():
+                if version in display_versions and value not in new_versions.values():
+                    new_versions[value.replace(".", "").replace("-", "")] = value
```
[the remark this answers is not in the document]
RobertKrawitz: Not a big deal; you can do regexp stuff in python too.

benchmark_runner/grafana/perf/extract_dashboard.py
RobertKrawitz: I mean if EOT is somehow present inside a line as opposed to being part of a here document.

RobertKrawitz (changes requested):
Just a few more quoting changes and one Content-Type:

.github/workflows/Nightly_Perf_Env_CI.yml
ebattat: Fixed

benchmark_runner/grafana/perf/extract_dashboard.py
ebattat: That generates by Terraform it also contains a strange prefix, this script do the work

ebattat:
@RobertKrawitz, The parser works fine and parse it without EOT

RobertKrawitz:
> @RobertKrawitz, The parser works fine and parse it without EOT

Just because it works in a test doesn't mean that it's robust.  You need to handle edge cases and unexpected input too.

ebattat:
@RobertKrawitz here is an example for the code we parse

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

```
Outputs:

dashboard = <<EOT
{
   "description": "Dashboard for Faro",
   "graphTooltip": 1,
   "panels": [
      {
         "datasource": {
            "type": "datasource",
            "uid": "-- Mixed --"
         },
         "fieldConfig": {
            "defaults": {
               "unit": "reqps"
            }
         },
         "gridPos": {
            "h": 8,
            "w": 24
         },
         "id": 1,
         "targets": [
            {
               "datasource": {
                  "type": "prometheus",
                  "uid": "ops-cortex"
               },
               "expr": "sum by (status_code) (rate(request_duration_seconds_count{job=~\".*/faro-api\"}[$__rate_interval]))"
            }
         ],
         "title": "Requests / sec",
         "type": "timeseries"
      }
   ],
   "schemaVersion": 36,
   "time": {
      "from": "now-6h",
      "to": "now"
   },
   "timezone": "utc",
   "title": "Faro dashboard"
}

EOT
```

RobertKrawitz:
What happens if the original is:

```
{
   "description": "Dashboard for Faro",
   "graphTooltip": 1,
   "panels": [
      {
         "datasource": {
            "type": "datasource",
            "uid": "-- Mixed --"
         },
         "fieldConfig": {
            "defaults": {
               "unit": "reqps"
            }
         },
         "gridPos": {
            "h": 8,
            "w": 24
         },
         "id": 1,
         "targets": [
            {
               "datasource": {
                  "type": "prometheus",
                  "uid": "EOT-cortex"
               },
               "expr": "sum by (status_code) (rate(request_duration_seconds_count{job=~\".*/faro-api\"}[$__rate_interval]))"
            }
         ],
         "title": "Requests / sec",
         "type": "timeseries"
      }
   ],
   "schemaVersion": 36,
   "time": {
      "from": "now-6h",
      "to": "now"
   },
   "timezone": "utc",
   "title": "Faro dashboard"
}
```

You'll then get
```
{
   "description": "Dashboard for Faro",
   "graphTooltip": 1,
   "panels": [
      {
         "datasource": {
            "type": "datasource",
            "uid": "-- Mixed --"
         },
         "fieldConfig": {
            "defaults": {
               "unit": "reqps"
            }
         },
         "gridPos": {
            "h": 8,
            "w": 24
         },
         "id": 1,
         "targets": [
            {
               "datasource": {
                  "type": "prometheus",
                  "uid": "
```

as your output.

I know this is unlikely, but the code needs to be written defensively.

ebattat:
@RobertKrawitz. so what do you recommend ?

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -314,3 +315,84 @@ jobs:
...
+
+      - name: Update latest product versions in main.libsonnet
+        env:
+          MAIN_LIBSONNET_PATH: ${{ secrets.PERF_MAIN_LIBSONNET_PATH }}
+          ELASTICSEARCH: ${{ secrets.PERF_ELASTICSEARCH }}
+          ELASTICSEARCH_PORT: ${{ secrets.PERF_ELASTICSEARCH_PORT }}
+          ELASTICSEARCH_USER: ${{ secrets.PERF_ELASTICSEARCH_USER }}
+          ELASTICSEARCH_PASSWORD: ${{ secrets.PERF_ELASTICSEARCH_PASSWORD }}
+        run: |
+          python $GITHUB_WORKSPACE/grafonnet_generator/grafana/update_versions_main_libsonnet.py
```
RobertKrawitz, line 340: Needs to be quoted.

.github/workflows/Perf_Env_Build_Test_CI.yml
```diff
@@ -261,9 +264,89 @@ jobs:
...
+
+      - name: Update latest product versions in main.libsonnet
+        env:
+          MAIN_LIBSONNET_PATH: ${{ secrets.PERF_MAIN_LIBSONNET_PATH }}
+          ELASTICSEARCH: ${{ secrets.PERF_ELASTICSEARCH }}
+          ELASTICSEARCH_PORT: ${{ secrets.PERF_ELASTICSEARCH_PORT }}
+          ELASTICSEARCH_USER: ${{ secrets.PERF_ELASTICSEARCH_USER }}
+          ELASTICSEARCH_PASSWORD: ${{ secrets.PERF_ELASTICSEARCH_PASSWORD }}
+        run: |
+          python $GITHUB_WORKSPACE/grafonnet_generator/grafana/update_versions_main_libsonnet.py
```
RobertKrawitz, line 287: Needs to be quoted.

RobertKrawitz (changes requested):
Few more quotes, and still need to fix the EOT regex.

.github/workflows/Nightly_Perf_Env_CI.yml
ebattat: Fixed

.github/workflows/Perf_Env_Build_Test_CI.yml
ebattat: Fixed

benchmark_runner/common/grafana/grafana_operations.py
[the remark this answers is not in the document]
ebattat: The Content-Type header is typically used when sending data in a request (e.g., in a POST or PUT request) to specify the format of the data being sent. Since I am only fetching data using a GET request and not sending any data, I don't need to include the Content-Type header in this case.
You can see that I use content type json={"dashboard": self.dashboard_data} in override_dashboard() below

ebattat:
@RobertKrawitz 

1. Http content type:
The Content-Type header is typically used when sending data in a request (e.g., in a POST or PUT request) to specify the format of the data being sent. Since I am only fetching data using a GET request and not sending any data, I don't need to include the Content-Type header in this case.
You can see that I use content type json={"dashboard": self.dashboard_data} in override_dashboard() method below

2. EOT
I have added dictionary validation by load it to json, meaning if we have the following example, we will get the following error: Extracted content is not valid JSON.

```
Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

Outputs:

dashboard = <<EOT
{
   "description": "Dashboard for Faro",
   "graphTooltip": "EOT",
   "title": "Faro dashboard"
}

EOT
"""
```

RobertKrawitz:
> @RobertKrawitz
> 
> 1. Http content type:
>    The Content-Type header is typically used when sending data in a request (e.g., in a POST or PUT request) to specify the format of the data being sent. Since I am only fetching data using a GET request and not sending any data, I don't need to include the Content-Type header in this case.

OK.

> 2. EOT
>    I have added dictionary validation by load it to json, meaning if we have the following example, we will get the following error: Extracted content is not valid JSON.

What you should do is improve the regexp so that it correctly handles this situation, not raise an error because a string happened to be within the text.

ebattat:
@RobertKrawitz.

The only way to handle it is to inject a special string instead of EOT, e.g. `@@EOT@@`.
We are using existing terraform jsonnet parser and we cannot change it[
](https://registry.terraform.io/providers/alxrem/jsonnet/latest/docs)
So the current solution is the best we can do, if there will be any parsing issue, exception will be raised.

RobertKrawitz:
> @RobertKrawitz.
> 
> The only way to handle it is to inject a special string instead of EOT, e.g. `@@EOT@@`. We are using existing [terraform jsonnet parser code ](https://registry.terraform.io/providers/alxrem/jsonnet/latest/docs) that generate auto the following output syntax:
> 
> ```
> dashboard = <<EOT
> {
>    "description": "Dashboard for Faro",
>    "graphTooltip": "aaa",
>    "title": "Faro dashboard"
> }
> 
> EOT
> ```
> 
> So the current solution is the best we can do, if there will be any parsing issue, exception will be raised.

No, the regular expression can be more intelligent, to only recognize the second EOT on a line by itself.  Rather than

```
r'dashboard = <<EOT(.*?)EOT'
```

it can be something like

```
r`dashboard = <<EOT(.*\n)EOT($|\n)'
```
to pick up the final EOT on a line by itself, either at the very end or between two newlines.

RobertKrawitz:
The other way to do it is to pass it to a shell to parse, but then you have to trust that the code isn't going to do something it shouldn't.  Parsing the here doc is reasonable enough, just do it correctly.

benchmark_runner/grafana/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+    go install github.com/google/go-jsonnet/cmd/jsonnet@latest && \
+    go install -a github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb@latest && \
+    wget https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip && \
+    unzip terraform_${TERRAFORM_VERSION}_linux_amd64.zip && \
+    mv terraform /usr/local/bin/ && \
+    rm terraform_${TERRAFORM_VERSION}_linux_amd64.zip
+
+ENV PATH="/root/go/bin:${PATH}"
+
+CMD ["/bin/sh", "-c", "terraform init; make > dashboard.txt;  python3 extract_dashboard.py dashboard.txt dashboard.json; terraform destroy -auto-approve > /dev/null 2>&1"]
```
RobertKrawitz, line 18: I think there's a much easier way you can do this:

```
CMD ["/bin/sh", "-c", "terraform init; make |/bin/sh > dashboard.json; terraform destroy -auto-approve > /dev/null 2>&1"]
```
and then you won't need extract_dashboard.py at all.

ebattat: Tried, but its not working:

```
[root@6e8e49594c69 app]# make |/bin/sh > dashboard.json
/bin/sh: line 1: $'\E[0m\E[1mdata.jsonnet_file.dashboard:': command not found
```

ebattat: I have updated the extract code for suffix: **\nEOT\n**
`r'dashboard = <<EOT(.*?)\nEOT\n'`
That should solve the EOT issue.

RobertKrawitz: I think you need to run `terraform init --no-color`.

RobertKrawitz: From https://developer.hashicorp.com/terraform/cli/commands/init
```
-no-color Disable color codes in the command output.
```

You're much better off not trying to fight what terraform init is trying to do here.

RobertKrawitz: What exactly does `make` produce here?  Spaces between the variable name, the `=`, and the text are not valid shell, so maybe it's not producing what I think it is and that regex may be the only way to go.  But I'd like to see both the `Makefile` and the output.

ebattat: The `terraform init` is not related to the problem at all.
The problem is with the `terraform apply` inside Makefile.
This command converts jsonnet(grafonnet) code into json and its using [Jsonnet terraform rendered code](https://registry.terraform.io/providers/alxrem/jsonnet/latest/docs) that rendered it into constant output
```
dashboard = <<EOT
{
   "description": "Dashboard for Faro",
   "graphTooltip": "aaa",
   "title": "Faro dashboard"
}

EOT
```

ebattat:
```
.PHONY: apply
apply:
        @cd jsonnet/ && \
                jb install && \
                cd ../ && \
                terraform apply -auto-approve > dashboard.txt && \
                python3 extract_dashboard.py dashboard.txt dashboard.json
```
@RobertKrawitz, I put all inside Makefile, WDYT ?

RobertKrawitz: Might as well put it all in the makefile; I still think you should avoid that temporary file (dashboard.txt).

RobertKrawitz: Where does `/root/grafonnet` come from?

ebattat: terraform apply -auto-approve -json => generates corrupted json output

ebattat: /root/grafonnet/ => You should put your local path.

RobertKrawitz: Local path to what?

Can you include the generated JSON?

benchmark_runner/grafana/perf/extract_dashboard.py
```diff
@@ -0,0 +1,19 @@
+import re, sys, json
+
+if len(sys.argv) != 3:
+    raise ValueError('Usage: python extract_dictionary.py input_file output_file')
```
RobertKrawitz, line 4: It's called extract_dashboard.  But you really should have it read from stdin and write to stdout, without any temporary files.

RobertKrawitz (changes requested):

benchmark_runner/grafana/Dockerfile
ebattat: [dashboard.json](https://github.com/redhat-performance/benchmark-runner/blob/3ec88b4167cca890904c6211866d51277f1675fb/benchmark_runner/grafana/perf/dashboard.json)
I have pushed the changes, think its good enough for now

benchmark_runner/grafana/perf/extract_dashboard.py
ebattat: You right, remove temporary file, read directly from stdin

`terraform apply -auto-approve | python3 extract_dashboard.py dashboard.json`

benchmark_runner/grafana/perf/Makefile
```diff
@@ -0,0 +1,5 @@
+apply:
+	@cd jsonnet/ && \
+		jb install && \
+		cd ../ && \
+		terraform apply -auto-approve | python3 extract_dashboard.py dashboard.json
```
RobertKrawitz, line 5: It would be best for extract_dashboard.py to simply send its output to stdout (so you could easily e. g. pipe it through `jq` if desired), but I'll accept this, as it avoids the temporary file.

RobertKrawitz (approved):

benchmark_runner/grafana/perf/Makefile
RobertKrawitz: Yes, I approved it as such.  I still think that extract_dashboard should write to stdout rather than to a separate file; that would allow you to e. g. process it further through jq, but I've accepted it as it avoids the temporary file.

ebattat: Do you mean to dashboard.json file?
We must have this file because we use it to update Grafana dashboard in case of changed.

RobertKrawitz: Correct, but you could write it as

```
terraform apply -auto-approve | python3 extract_dashboard.py > dashboard.json
```

Then if you want at some point to process the JSON with jq downstream of extract_dashboard.py you can simply add it to the pipeline.

ebattat: You right, updated it !!!

RobertKrawitz (approved):

benchmark_runner/common/grafana/update_grafana_last_value_mappings.py
```diff
@@ -0,0 +1,116 @@
...
+
+from benchmark_runner.main.environment_variables import environment_variables
+from benchmark_runner.common.elasticsearch.elasticsearch_operations import ElasticSearchOperations
+
+
+class UpdateGrafanaLastValueMappings:
+    """
+    This class updates grafana dashboard with last value mappings from ElasticSearch
+    """
+    MAX_VERSION_LEN = 20
```
RobertKrawitz, line 13: What's this for (please document it).

ebattat: @RobertKrawitz, add document.
Prevent error messages from being displayed instead of versions in case of a connection issue

ebattat:
@RobertKrawitz,

I have added jsonnetfile.lock.json file for working with specific grafonnet and not the latest grafonnet version.
That should reduce failure while using grafonnet generator. 
I have added it due to recent grafonnet code [fix](https://github.com/grafana/grafonnet/commit/2359881e8a9e5601e3ccf8f7f2b024f06a70adf0) that cause a failure when generating dashboard.json. 
WDYT ?

benchmark_runner/grafana/perf/jsonnet/jsonnetfile.lock.json
```diff
@@ -0,0 +1,86 @@
+{
```
RobertKrawitz, line 1: Rather than `.lock`, I suggest `.version` if that's possible (i. e. if jsonnet doesn't have some kind of hard-coded requirement for that filename).  `lock` implies that it's locking against some kind of concurrency issue; this is about versioning, not concurrency.

RobertKrawitz (changes requested):
Unless this is required by jsonnet, please change the version file name from jsonnetfile.lock.json to jsonnetfile.version.json or similar.  `lock` implies concurrency control of some kind, which isn't what this is about.

benchmark_runner/grafana/perf/jsonnet/jsonnetfile.lock.json
RobertKrawitz: That looks fine.

RobertKrawitz (approved):

PR #651, opened 2023-09-06
update pypi version

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Update version due to last failure


RobertKrawitz (commented):
What does this actually change?  It looks like it bumps the version with no payload.

PR #654, opened 2023-09-10
Update Prometheus VM memory metrics

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. Adding **VM memory metrics** to bootstrom and vdbench workloads by running the following prometheus queries:
```
- query: container_memory_cache{pod=~"virt-launcher.*", container="compute"}
  metricName: VM Memory Cache

- query: container_memory_working_set_bytes{pod=~"virt-launcher.*", container="compute"}
  metricName: VM Memory working set bytes

- query: container_memory_max_usage_bytes{pod=~"virt-launcher.*", container="compute"}
  metricName: VM Memory Max Usage Bytes

```
2. Adding **VM memory metrics** to Grafana Dashboard by adding Grafonnet code in main.libsonnet.

### DIFF
--- a/benchmark_runner/common/prometheus/metrics-default.yaml
+++ b/benchmark_runner/common/prometheus/metrics-default.yaml
@@ -1,4 +1,5 @@
 
+
 # container CPU/Memory
 
 - query: (sum(irate(container_cpu_usage_seconds_total{container!="",namespace="benchmark-runner", mode!="idle"}[2m])) by (node)) > 0
@@ -7,4 +8,12 @@
 - query: sum(container_memory_rss{namespace="benchmark-runner"} and on (node) kube_node_role{role="worker"}) by (node) > 0
   metricName: containerMemory-benchmark-runner
 
+- query: container_memory_cache{pod=~"virt-launcher.*", container="compute"}
+  metricName: VM Memory Cache
+
+- query: container_memory_working_set_bytes{pod=~"virt-launcher.*", container="compute"}
+  metricName: VM Memory working set bytes
+
+- query: container_memory_max_usage_bytes{pod=~"virt-launcher.*", container="compute"}
+  metricName: VM Memory Max Usage Bytes
 
--- a/benchmark_runner/common/prometheus/prometheus_metrics_operations.py
+++ b/benchmark_runner/common/prometheus/prometheus_metrics_operations.py
@@ -8,17 +8,18 @@ from urllib3.exceptions import InsecureRequestWarning
 from urllib3 import disable_warnings
 disable_warnings(InsecureRequestWarning)
 
+from benchmark_runner.main.environment_variables import environment_variables
 from benchmark_runner.common.logger.logger_time_stamp import logger_time_stamp, logger
-from benchmark_runner.workloads.workloads_operations import WorkloadsOperations
 
 
-class PrometheusMetricsOperation(WorkloadsOperations):
+class PrometheusMetricsOperation:
     """
-    This class Run prometheus queries
+    This class contains methods for running prometheus queries, parsing the results into dictionary nad uploading it into elasticsearch
     """
 
     def __init__(self):
         super().__init__()
+        self._environment_variables_dict = environment_variables.environment_variables_dict
         self.__current_dir = os.path.dirname(os.path.abspath(__file__))
         self.__queries_file = os.path.join(self.__current_dir, 'metrics-default.yaml')
         self.__authorization = {'Authorization': self.__get_prometheus_token()}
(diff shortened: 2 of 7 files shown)


benchmark_runner/common/prometheus/metrics-default.yaml
```diff
@@ -7,4 +8,12 @@
 - query: sum(container_memory_rss{namespace="benchmark-runner"} and on (node) kube_node_role{role="worker"}) by (node) > 0
   metricName: containerMemory-benchmark-runner
 
+- query: container_memory_cache{pod=~"virt-launcher.*", container="compute"}
+  metricName: VM Memory Cache
```
RobertKrawitz, line 12: I'm not certain that spaces in metric names work correctly.  Make sure you test this.

benchmark_runner/common/prometheus/prometheus_metrics_operations.py
```diff
@@ -160,4 +164,33 @@ def upload_result_to_elastic(self):
                 print(query, results)
         else:
             raise Exception('Missing ElasticSearch data')
-        
+
+    @staticmethod
+    @logger_time_stamp
+    def parse_prometheus_metrics(data: dict):
+        """
+        This method parses prometheus metrics and return summary result
```
RobertKrawitz, line 172: returns

benchmark_runner/common/prometheus/prometheus_metrics_operations.py
```diff
@@ -160,4 +164,33 @@ def upload_result_to_elastic(self):
...
+        based on queries: /benchmark_runner/common/prometheus/metrics-default.yaml
+        @return:
+        """
+        result_dict = {}
+        for query, data_list in data.items():
+            if 'containerCPU-benchmark-runner' in query:
+                suffix = 'CPU'
```
RobertKrawitz, line 179: This needs to be kept in sync with the metric names.  Since those are under your control, this isn't a problem, just decide whether there might be other similar metrics that you want to do this with.

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -61,6 +60,33 @@ def _create_vm_scale(self, vm_num: str):
         self._oc.create_async(yaml=os.path.join(f'{self._run_artifacts_path}', f'{self._name}_{vm_num}.yaml'))
         self._oc.wait_for_vm_status(vm_name=f'{self._workload_name}-{self._trunc_uuid}-{vm_num}', status=VMStatus.Stopped)
 
+    def _run_vm(self):
+        """
+        This method run vm
```
RobertKrawitz, line 65: runs

This comment isn't helpful as it simply duplicates the name of the method; there's no point in even having it here at all.  If you want to explain it here do so, but a comment "because there's supposed to be a comment on each method" is gratuitous.

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -24,7 +23,7 @@ def __init__(self):
         self._vm_name = ''
         self._data_dict = {}
         self._bootstorm_start_time = {}
-        self._prometheus_metrics_operation = PrometheusMetricsOperation()
+
```
RobertKrawitz, line 26: I believe there's supposed to be only one blank line between method definitions (2 lines at top level).

RobertKrawitz (commented):
Nothing major.

benchmark_runner/common/prometheus/metrics-default.yaml
ebattat: Yes, works for me in Grafana and its more readable than VM-Memory-Cache

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -108,45 +133,34 @@ def _wait_for_delete_vm_scale(self, vm_num: str):
         """
         self._oc.wait_for_vm_delete(vm_name=f'{self._workload_name}-{self._trunc_uuid}-{vm_num}')
 
+    def _initialize_run(self):
+        """
+        Initialize run parameters
+        @return:
```
RobertKrawitz, line 139: This is another comment that either needs to be expanded on or removed.  It's actually not really accurate, since it doesn't just initialize parameters, it also initializes prometheus and more importantly has a side effect, creating the namespace.

ebattat: Updated, pls let me know if it more clear

RobertKrawitz (approved):

PR #660, opened 2023-09-13
install latest 4.14.0-rc

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Install the following:
1. latest [4.14.0-rc](https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/releasestream/4-stable/release/4.14.0-rc.0)
2. CNV_VERSION: "4.14"
3. ODF_VERSION: "4.13"
4. oc_version="4.13.0-0.okd-2023-06-04-080300"
5.  virtctl_version="1.0.0"

### DIFF
--- a/.github/workflows/Perf_Env_Build_Test_CI.yml
+++ b/.github/workflows/Perf_Env_Build_Test_CI.yml
@@ -67,13 +67,13 @@ jobs:
           IBM_BUCKET: ${{ secrets.IBM_BUCKET }}
           IBM_KEY: ${{ secrets.IBM_KEY }}
           RUN_ARTIFACTS_URL: ${{ secrets.PERF_RUN_ARTIFACTS_URL }}
+          OC_CLIENT_VERSION: ${{ secrets.PERF_OC_CLIENT_VERSION }}
         run: |
           # Install Dockerfile content for pytest
           # install oc/kubectl
-          oc_version="4.13.0-0.okd-2023-06-04-080300"
-          curl -L "https://github.com/openshift/okd/releases/download/${oc_version}/openshift-client-linux-${oc_version}.tar.gz" -o "$RUNNER_PATH/openshift-client-linux-${oc_version}.tar.gz"
-          tar -xzvf $RUNNER_PATH/openshift-client-linux-${oc_version}.tar.gz -C $RUNNER_PATH/
-          rm $RUNNER_PATH/openshift-client-linux-${oc_version}.tar.gz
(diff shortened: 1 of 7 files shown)


.github/workflows/Perf_Env_Build_Test_CI.yml
```diff
@@ -70,7 +70,7 @@ jobs:
         run: |
           # Install Dockerfile content for pytest
           # install oc/kubectl
-          oc_version="4.13.0-0.okd-2023-06-04-080300"
+          oc_version="4.13.0-0.okd-2023-09-03-082426"
```
RobertKrawitz, line 73: For the future, it would be really good to parameterize this somewhere so you don't need to modify so many places.  Possibly this should be a secret or otherwise something in the environment.  But it is important to have a PR when these change, so we can determine when things change.

RobertKrawitz (approved):
As I noted in my one comment, these versions should be parameterized so it isn't necessary to change so many places.  That's not sufficient reason not to approve the PR, but please consider how to do this.

.github/workflows/Perf_Env_Build_Test_CI.yml
ebattat: We can put it in secret but it will not be visible for us, do you think its ok ?

RobertKrawitz: A secret's not really right here because we can't determine when it changes.  Can we template these versions?

RobertKrawitz: I'm just not familiar with how github actions work to know whether we can provide template variables in a command line or something.

jeniferh (approved):
Agreed, changes look fine for now, but I can't recall if there is a reason we don't just pull the oc client from the mirror where the version of the installer is stored? for ex:
https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.14.0-rc.0/
has openshift-client-linux tar balls

ebattat:
@jeniferh, I have updated the client path: "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/${oc_version}/openshift-client-linux.tar.gz" in all the places.

jeniferh:
Thanks, LGTM. Much cleaner that way.

ebattat:
@RobertKrawitz, I have updated secret for OC_CLIENT_VERSION, except 1 place that I remain the exact oc version.
Please review it.

PR #662, opened 2023-09-13
fix pull push grafana json

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Still getting pull/push [error](https://github.com/redhat-performance/benchmark-runner/actions/runs/6169715109/job/16746831451)
Fix pull/push Grafana json & main.libsonnet file in case of change same as [bump version working example](https://github.com/redhat-performance/benchmark-runner/blob/9f3813f983d433adff778a3d5913d1f80ec95d14/.github/workflows/Perf_Env_Build_Test_CI.yml#L284-L297)


.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -378,14 +378,13 @@ jobs:
           git checkout main
           git config --global user.email "${{ secrets.EMAIL }}"
           git config --global user.name  "${{ secrets.USER_NAME }}"
+          git config pull.rebase false  # merge (the default strategy)
```
RobertKrawitz, line 381: Why does rebase not work correctly?

.github/workflows/Nightly_Perf_Env_CI.yml
```diff
@@ -378,14 +378,13 @@ jobs:
...
           git add "$GITHUB_WORKSPACE"/benchmark_runner/grafana/perf/dashboard.json
           git commit -m "Update grafana json file"
 
-          # Pull the latest changes from the remote main branch
-          git pull origin main
-
-          # Push the changes to the remote main branch
-          git push origin main
+          git pull https://${{ secrets.GIT_TOKEN }}@${{ secrets.GIT_REPOSITORY}} main
+          git push https://${{ secrets.GIT_TOKEN }}@${{ secrets.GIT_REPOSITORY}} main
```
RobertKrawitz, line 387: What is the issue here?

.github/workflows/Nightly_Perf_Env_CI.yml
ebattat: We need to add it according to the error: https://github.com/redhat-performance/benchmark-runner/actions/runs/6169715109/job/16746831451

.github/workflows/Nightly_Perf_Env_CI.yml
ebattat: no special issue, prefer to use it as we use in bumb version: https://github.com/redhat-performance/benchmark-runner/blob/9f3813f983d433adff778a3d5913d1f80ec95d14/.github/workflows/Perf_Env_Build_Test_CI.yml#L284-L297

.github/workflows/Nightly_Perf_Env_CI.yml
RobertKrawitz: That looks like we simply weren't specifying any strategy at all, and we need to specify something.  Why would rebase or fast-forward not work correctly?

.github/workflows/Nightly_Perf_Env_CI.yml
RobertKrawitz: That seems unnecessary; isn't origin already correct?

.github/workflows/Nightly_Perf_Env_CI.yml
ebattat: No because we need to merge

.github/workflows/Nightly_Perf_Env_CI.yml
ebattat: I know that its working.
I can try:

```
git pull main
git push main
```
no need origin, we update the upstream.

.github/workflows/Nightly_Perf_Env_CI.yml
RobertKrawitz: So what exactly needs to be merged (as opposed to rebased)?

.github/workflows/Nightly_Perf_Env_CI.yml
RobertKrawitz: OK, it's just easier to read without all that extra.

.github/workflows/Nightly_Perf_Env_CI.yml
ebattat: `hint:   git config pull.rebase false  # merge`

RobertKrawitz: It only says that you have to use one of the three strategies.

For that matter, why couldn't you do

```
git pull main
git add "$GITHUB_WORKSPACE"/benchmark_runner/grafana/perf/dashboard.json
git commit -m "Update grafana json file"
git push main
```
rather than having to do any kind of merge?

ebattat: Because we are merging, I remembered I got the same error with bump version.

RobertKrawitz: Merging what?  It looks like you're just updating dashboard.json, so you should be able to do

```
git checkout main
git pull
git add "$GITHUB_WORKSPACE"/benchmark_runner/grafana/perf/dashboard.json
git commit -m "Update grafana json file"
git push
```

RobertKrawitz (changes requested):

.github/workflows/Nightly_Perf_Env_CI.yml
ebattat: @RobertKrawitz, 
It will not work, we have already tested it like this and it failed

RobertKrawitz: That's what I'm trying to understand.  How did it fail?  It looks to me like you're not merging the file, you're just updating it.

ebattat: @RobertKrawitz We are merging the json file into main.
This is the only option to make it works!

RobertKrawitz: I think I see, the push is to the actual repository, so you can't rebase it.  OK, approving.

RobertKrawitz (approved):

PR #672, opened 2023-09-21
fix OCP 4.14 LSO issue

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. OpenShift 4.14 PV disk uses the 'wwn-' prefix instead of 'scsi-'. 
2. Solve it by removing the 'wwn-' prefix and replace it with 'scsi-' for the LSO free disk.
3. Fix pytest because method return string id and not list.
Thats the reason for PostgreSQL LSO [failure](https://github.com/redhat-performance/benchmark-runner/actions/runs/6222650769/job/16887021038) in PerfCI.


benchmark_runner/common/oc/oc.py
```diff
@@ -88,12 +84,13 @@ def remove_lso_path(self):
...
+        :return: It returns the free disk ID string with the 'scsi-3600062b20' prefix [add '20' because it is missing in wwn-]
         """
-        workers_disk_ids = self.get_worker_disk_ids()
-        workers_pv_disk_ids = self.get_pv_disk_ids()
-        if workers_disk_ids:
-            return list(set(workers_disk_ids) - set(workers_pv_disk_ids))
+        workers_disk_ids = [disk_id.lstrip('wwn-0x600062b').lstrip('scsi-3600062b') for disk_id in self.get_worker_disk_ids()]
+        workers_pv_disk_ids = [disk_id.lstrip('wwn-0x600062b').lstrip('scsi-3600062b') for disk_id in self.get_pv_disk_ids()]
+        if self.get_worker_disk_ids():
+            return f'scsi-3600062b20{list(set(workers_disk_ids) - set(workers_pv_disk_ids))[0]}'
```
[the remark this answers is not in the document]
ebattat: I have updated the code, please review it.
You right no need '20' anymore.
BTW, I have updated SCSI_PREFIX and WWN_PREFIX as constant:
```
SCSI_PREFIX = 'scsi-3600062b'
WWN_PREFIX = 'wwn-0x600062b'
```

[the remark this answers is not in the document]
ebattat: @RobertKrawitz, You right.
I have updated the code to be more generic, pls let me know if you have more comments.

benchmark_runner/common/oc/oc.py
```diff
@@ -27,6 +27,7 @@ class OC(SSH):
     SHORT_TIMEOUT = 600
     # sleep time between checks is 5 sec
     SLEEP_TIME = 3
+    DISK_ID_PREFIX_LEN = 13
```
RobertKrawitz, line 30: Where does the prefix length come from?  Why can't you use the entire string?

ebattat: No need to specify entry string anymore. The parsing of every storage id scsi or wwn is the same.

RobertKrawitz: I don't understand why you need a fixed prefix length (or any kind of prefix) here.

RobertKrawitz: Put another way, why aren't you just using the string without interpretation?

ebattat: The main problem is that ODF is using **wwn-** prefix and LSO using **scsi-** prefix.
So I added this code to workaround the disk id type wwn\scsi prefix.

RobertKrawitz: Right, but other than the 'scsi-3` vs. `wwn-0x` (6 characters rather than 13), the names are identical.

What I don't understand is why you don't simply use the full name.  If you need to use wwn- names for ODF, just use the full wwn- name.

ebattat: To extract the unique id from the storage, need to cut 13 characters.
**scsi-3600062b**2070785c028ae89f244909b6c
**wwn-0x600062b**2070785c028ae89f244909b6c

I use 'scsi-' for LSO disk.
This should resolve the issue in the long term, even if they revert ODF back to using 'scsi-' instead of 'wwn-'

ebattat: ODF PV changed the disk format from **scsi-** to **wwn-** in OCP 4.14
So in order to find the free disk we should 
subtraction all_workers_disk[scsi- format] - odf_worker_disk[wwn- format] 
Thats the reason I remove the storage type in both.

RobertKrawitz: I understand you need to remove the `scsi-3` and `wwn-0x` prefixes, but the other 7 bytes are not part of the prefix, they are part of the ID.

RobertKrawitz: For example, on various of my systems I have disks named

```
wwn-0x5000c500ac5f71a3
scsi-35000c500ac5f71a3

wwn-0x6d094660593af20022eb5a6eb8838bae
scsi-36d094660593af20022eb5a6eb8838bae

wwn-0x55cd2e414f75a670
scsi-355cd2e414f75a670
```

Note that the lengths of these names differ.  You cannot make any assumptions beyond the first 6 bytes.

ebattat: So what you recommend to do ?

RobertKrawitz: Use the entire string except for the first 6 bytes.

benchmark_runner/common/oc/oc.py
```diff
@@ -27,6 +27,7 @@ class OC(SSH):
     SHORT_TIMEOUT = 600
     # sleep time between checks is 5 sec
     SLEEP_TIME = 3
+    DISK_ID_PREFIX_LEN = 6
```
RobertKrawitz, line 30: No longer need this.

benchmark_runner/common/oc/oc.py
```diff
@@ -68,32 +70,39 @@ def get_pv_disk_ids(self):
...
-        The method removes lso path on each node
-        @return:
-        """
-        self.run(fr"""{self.__cli} get nodes -l node-role.kubernetes.io/worker= -o jsonpath="{{range .items[*]}}{{.metadata.name}}{{'\\n'}}{{end}}" |  xargs -I{{}} oc debug node/{{}} -- chroot /host sh -c "rm -rf /mnt/local-storage/local-sc" """)
-
     def get_free_disk_id(self):
         """
-        This method returns free disk (workers_all_disk_ids - workers_odf_pv_disk_ids)
+        This method extracts the free disk by worker_disk_ids - pv_disk_ids
+        :return: It returns the first free disk ID with the 'scsi-3600062b' prefix
```
RobertKrawitz, line 92: This needs to be updated.

benchmark_runner/common/oc/oc.py
```diff
@@ -68,32 +70,39 @@ def get_pv_disk_ids(self):
...
-        workers_disk_ids = self.get_worker_disk_ids()
-        workers_pv_disk_ids = self.get_pv_disk_ids()
-        if workers_disk_ids:
-            return list(set(workers_disk_ids) - set(workers_pv_disk_ids))
+        free_disk_id = ''
+        if self.get_worker_disk_ids():
+            workers_disk_ids = [disk_id for disk_id in self.get_worker_disk_ids()]
+            workers_pv_disk_ids = [disk_id[self.DISK_ID_PREFIX_LEN:] for disk_id in self.get_pv_disk_ids()]
+            # Calculate the set difference
+            unique_disk_ids = list(set(workers_disk_ids) - set(workers_pv_disk_ids))
```
RobertKrawitz, line 99: Rather than this, you need something like

```
unique_disk_ids = []
for worker_disk in workers_disk_ids:
    disk_matched = False
    for pv_disk in workers_pv_disk_ids:
        if workers_disk.endsWith(pv_disk):
            disk_matched=True
            break
    if not disk_matched:
        unique_disk_ids.append(prefix + worker_disk
```

or some such.

RobertKrawitz (changes requested):

benchmark_runner/common/oc/oc.py
RobertKrawitz: Note that when the prefix was `google-` it was 7 bytes, rather than 6.  You cannot make assumptions about the prefix!

RobertKrawitz: You don't need to.  All you need to do is write the code to be independent of type, by simply matching the unprefixed ID against the list of IDs by using `endsWith`

RobertKrawitz (approved):

PR #682, opened 2023-10-10
LSO workloads: Fix get free disk id

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Simplify the process of extracting the free disk ID by removing the disk ID prefix and using only the ID to find the free disk

### DIFF
--- a/benchmark_runner/benchmark_operator/benchmark_operator_workloads_operations.py
+++ b/benchmark_runner/benchmark_operator/benchmark_operator_workloads_operations.py
@@ -74,8 +74,8 @@ class BenchmarkOperatorWorkloadsOperations:
         if self._enable_prometheus_snapshot:
             self._snapshot = PrometheusSnapshot(oc=self._oc, artifacts_path=self._run_artifacts_path, verbose=True)
         self._prometheus_metrics_operation = PrometheusMetricsOperation()
-        # update lso_disk_id
-        if self._oc.get_free_disk_id():
+        # Update lso_disk_id only if both worker_disk_ids and a free disk exist
+        if self._environment_variables_dict.get('worker_disk_ids', '') and self._oc.get_free_disk_id():
             self._lso_disk_id = self._oc.get_free_disk_id()
             self._environment_variables_dict['lso_disk_id'] = self._lso_disk_id
         else:
--- a/benchmark_runner/common/oc/oc.py
+++ b/benchmark_runner/common/oc/oc.py
@@ -62,13 +62,6 @@ class OC(SSH):
         """
         return self.run(f"{self.__cli} get csv -n openshift-storage -ojsonpath='{{.items[0].spec.labels.full_version}}'")
 
-    def get_pv_disk_ids(self):
-        """
-        This method returns list of pv disk ids
-        """
-        result = self.run(f"{self.__cli} get pv -o jsonpath={{.items[*].metadata.annotations.'storage\.openshift\.com/device-id'}}")
-        return result.split()
-
     def remove_lso_path(self):
         """
         The method removes lso path on each node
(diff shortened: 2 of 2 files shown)


benchmark_runner/common/oc/oc.py
```diff
@@ -82,26 +75,29 @@ def get_worker_disk_ids(self):
...
-                free_disk_id = f'{self.__worker_disk_prefix}{worker_disk_id}'
-                break  # Exit the loop once a free disk ID is found
-
-        return free_disk_id
+        This method returns free disk (workers_all_disk_ids - workers_odf_pv_disk_ids) with its prefix
+        """
+        workers_disk_ids = self.get_worker_disk_ids()
+        workers_pv_disk_ids = self.get_pv_disk_ids()
+        if workers_disk_ids:
+            return f'{self.__worker_disk_prefix}{list(set(workers_disk_ids) - set(workers_pv_disk_ids))[0]}'
```
RobertKrawitz, line 100: Raise an exception if no disks are found.

benchmark_runner/common/oc/oc.py
```diff
@@ -82,26 +75,29 @@ def get_worker_disk_ids(self):
         """
         workers_disk_ids = []
         if self.__worker_disk_ids:
-            workers_disk_ids = [disk_id for disk_list in self.__worker_disk_ids.values() for disk_id in disk_list]
+            for node, disk_ids in self.__worker_disk_ids.items():
+                for disk_id in disk_ids:
+                    workers_disk_ids.append(disk_id)
```
RobertKrawitz, line 80: Why are you changing this?  And you should raise an exception if there's nothing.

```
if self.__worker_disk_ids:
    return [disk_id for disk_list in self.__worker_disk_ids.values() for disk_id in disk_list]
else
    raise ...
```

benchmark_runner/common/oc/oc.py
```diff
@@ -82,26 +75,29 @@ def get_worker_disk_ids(self):
...
         return workers_disk_ids
 
+    def get_pv_disk_ids(self):
+        """
+        This method returns list of pv disk ids
+        """
+        result = []
+        pvs = self.run(f"{self.__cli} get pv -o jsonpath={{.items[*].metadata.annotations.'storage\.openshift\.com/device-id'}}")
+        for pv in pvs.split():
+            result.append(pv[len(self.__worker_disk_prefix):])
```
RobertKrawitz, line 90: Remove the prefix, something like

```
return [ pv.lstrip(self.__worker_disk_prefix) for pv in pvs.split() if pv.startswith(self.__worker_disk_prefix)]
```

RobertKrawitz (changes requested):

benchmark_runner/benchmark_operator/benchmark_operator_workloads_operations.py
```diff
@@ -74,8 +74,8 @@ def __init__(self):
         if self._enable_prometheus_snapshot:
             self._snapshot = PrometheusSnapshot(oc=self._oc, artifacts_path=self._run_artifacts_path, verbose=True)
         self._prometheus_metrics_operation = PrometheusMetricsOperation()
-        # update lso_disk_id
-        if self._oc.get_free_disk_id():
+        # Update lso_disk_id only if worker_disk_ids exist
+        if self._environment_variables_dict.get('worker_disk_ids', ''):
             self._lso_disk_id = self._oc.get_free_disk_id()
```
RobertKrawitz, line 79: You need to make sure that you actually find a disk here -- it's all well and good for the secrets to specify a disk ID, but you need to make sure that you can find one at runtime.

benchmark_runner/common/oc/oc.py
RobertKrawitz: Again, make sure that you actually find a disk ID here.

benchmark_runner/common/oc/oc.py
```diff
@@ -82,26 +75,29 @@ def get_worker_disk_ids(self):
...
+                for disk_id in disk_ids:
+                    workers_disk_ids.append(disk_id)
         return workers_disk_ids
 
+    def get_pv_disk_ids(self):
+        """
+        This method returns list of pv disk ids
+        """
+        pv_ids = self.run(f"{self.__cli} get pv -o jsonpath={{.items[*].metadata.annotations.'storage\.openshift\.com/device-id'}}")
+        return [pv[len(self.__worker_disk_prefix):] for pv in pv_ids.split()]
```
RobertKrawitz, line 88: Actually strip off the prefix (not just by length, by content).

benchmark_runner/common/oc/oc.py
```diff
@@ -82,26 +75,29 @@ def get_worker_disk_ids(self):
...
-                free_disk_id = f'{self.__worker_disk_prefix}{worker_disk_id}'
-                break  # Exit the loop once a free disk ID is found
-
-        return free_disk_id
+        This method returns free disk (workers_all_disk_ids - workers_odf_pv_disk_ids) with its prefix
+        """
+        workers_disk_ids = self.get_worker_disk_ids()
+        workers_pv_disk_ids = self.get_pv_disk_ids()
+        free_disk_id = f'{self.__worker_disk_prefix}{list(set(workers_disk_ids) - set(workers_pv_disk_ids))[0]}'
+        if free_disk_id:
```
RobertKrawitz, line 97: The current code will always return something, because the prefix needs to be prepended only if you find a disk.

```
free_disk_id = {list(set(workers_disk_ids) - set(workers_pv_disk_ids))[0]}'
if free_disk_id:
    return f'{self.__worker_disk_prefix}{free_disk_id}'
else:
    raise Exception("No free disk found")
```

RobertKrawitz (changes requested):

benchmark_runner/benchmark_operator/benchmark_operator_workloads_operations.py
ebattat: add self._oc.get_free_disk_id()

RobertKrawitz: What happens if `self._oc_get_free_disk_id()` does not find a suitable disk?

RobertKrawitz: OK.  I think it should raise an exception if it doesn't get a free disk ID (this is a real problem), but however you want to handle it.  What happens if, for example, the only remaining disk goes offline?

benchmark_runner/common/oc/oc.py
RobertKrawitz: OK

benchmark_runner/common/oc/oc.py
RobertKrawitz: Again, don't remove the length of the prefix, actually strip the prefix.

RobertKrawitz (changes requested):

benchmark_runner/benchmark_operator/benchmark_operator_workloads_operations.py
ebattat: Fixed

ebattat:
I must use the length because we dont know if it 'scsi-3' or 'wwn-0x', but we know that the length of both is the same.
Currently ODF is using 'wwn-0x' while LSO is using 'scsi-3'.

RobertKrawitz:
> @RobertKrawitz, I must use the length because we don't know whether it's 'scsi-3' or 'wwn-0x,' but we know that the length of both is the same. Currently, ODF uses 'wwn-0x,' while LSO uses 'scsi-3.

Understood, but make sure to comment this very well!

RobertKrawitz:
```
# SATA disks in /dev/disk/by-id conventionally are prefixed with `wwn-0x` and `scsi-3`; ODF requires use of the `wwn-0x` prefix while LSO requires the `scsi-3` prefix.  We have to be prepared to use either prefix, but as they are the same length, we can presently strip the first 6 characters of the name.
```

benchmark_runner/common/oc/oc.py
ebattat: @RobertKrawitz updated

benchmark_runner/common/oc/oc.py
RobertKrawitz: This really should raise an exception if nothing is found.  This is expected to find something; if it doesn't, it indicates that something is wrong here.

RobertKrawitz (approved):

PR #687, opened 2023-10-29
Fix delete disk

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Got the following error while ODF installation
`Message:      failed to configure devices: failed to get device already provisioned by ceph-volume raw: osd.8:`
Fix delete ODF/ Ceph disks according to [recommendation](https://rook.io/docs/rook/v1.12/Getting-Started/ceph-teardown/#delete-the-data-on-hosts)

### DIFF
--- a/benchmark_runner/common/ocp_resources/create_ocp_resource.py
+++ b/benchmark_runner/common/ocp_resources/create_ocp_resource.py
@@ -77,23 +77,23 @@ class CreateOCPResource:
         self.remove_resource_files(path=os.path.join(self.__dir_path, resource))
         resource_files = self.get_sorted_resources(resource=resource)
         if 'lso' == resource:
-            self.__create_lso = CreateLSO(self.__oc, path=os.path.join(self.__dir_path, resource), resource_list=resource_files)
-            self.__create_lso.create_lso()
+            create_lso = CreateLSO(self.__oc, path=os.path.join(self.__dir_path, resource), resource_list=resource_files)
+            create_lso.create_lso()
         elif 'odf' == resource:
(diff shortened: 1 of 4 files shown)


benchmark_runner/common/ocp_resources/create_ocp_resource.py
```diff
@@ -77,23 +77,23 @@ def create_resource(self, resource: str):
         self.remove_resource_files(path=os.path.join(self.__dir_path, resource))
         resource_files = self.get_sorted_resources(resource=resource)
         if 'lso' == resource:
-            self.__create_lso = CreateLSO(self.__oc, path=os.path.join(self.__dir_path, resource), resource_list=resource_files)
-            self.__create_lso.create_lso()
+            create_lso = CreateLSO(self.__oc, path=os.path.join(self.__dir_path, resource), resource_list=resource_files)
+            create_lso.create_lso()
```
RobertKrawitz, line 81: What's the issue with this?

benchmark_runner/common/ocp_resources/odf/template/01_delete_disks_template.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env sh
+
+while (($# >= 2)) ; do
+    node=$1
+    delete_disk_command=$2
```
RobertKrawitz, line 5: Should be "=$2" (in quotes) so the shell won't process the value of $2.

ebattat: Fixed

benchmark_runner/common/ocp_resources/create_ocp_resource.py
ebattat: No need to create class object if you are not using it in several places

PR #690, opened 2023-11-04
Fix ODF installation

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Fix ODF installation:
1. Add sleep 30 sec between each ODF yaml for avoiding ODF installation failure
2. odf_disk_count: display -1 when 0,1 for avoiding conflict with 0/1 status code
3. Remove unused/ duplicated method: _install_and_wait_for_resource [duplicated: [__install_and_wait_for_resource_with_retry](https://github.com/redhat-performance/benchmark-runner/blob/3e85b76f3dbaab87ff529da685927e6bcd91e8c4/benchmark_runner/common/ocp_resources/create_kata.py#L23C4-L42)]

### DIFF
--- a/benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
+++ b/benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
@@ -10,7 +10,7 @@ from benchmark_runner.common.ocp_resources.create_ocp_resource_exceptions import
 
 class CreateOCPResourceOperations:
     """
-    This class is create OCP resources
+    This class is created OCP resources
     """
     def __init__(self, oc: OC):
         self._environment_variables_dict = environment_variables.environment_variables_dict
@@ -20,7 +20,7 @@ class CreateOCPResourceOperations:
     @typechecked
     def _replace_in_file(file_path: str, old_value: str, new_value: str):
         """
-        This method replace string in file
+        This method replaces string in file
         :param file_path:
         :param old_value:
         :param new_value:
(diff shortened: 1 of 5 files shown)


benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
```diff
@@ -10,7 +10,7 @@
 
 class CreateOCPResourceOperations:
     """
-    This class is create OCP resources
+    This class is created OCP resources
```
RobertKrawitz, line 13: This class creates

benchmark_runner/common/ocp_resources/create_odf.py
```diff
@@ -69,4 +71,7 @@ def create_odf(self):
                     self.wait_for_ocp_resource_create(resource='odf',
                                                       verify_cmd='oc get pod -n openshift-storage | grep osd | grep -v prepare | wc -l',
                                                       count_openshift_storage=True)
+            # sleep between each resource run for avoiding installation failure
+            logger.info(f"sleep {self._environment_variables_dict.get('bulk_sleep_time', '')} seconds")
+            time.sleep(int(self._environment_variables_dict.get('bulk_sleep_time', '')))
```
RobertKrawitz, line 76: Can you wait on a condition instead?  In code removed above there was a comment explaining why you can't wait on a condition in that case; if that's the situation here, add that comment.

RobertKrawitz (changes requested):

benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
ebattat: I will fix it in next PR

benchmark_runner/common/ocp_resources/create_odf.py
ebattat: I dont know if that the reason that ODF installation failed. 
Try to add this sleep time and see if Installation will pass successfully 
If not I will revert it back !!!

ebattat:
@RobertKrawitz, PerfCi is down due to this issue.
It is working for me when run it manually locally.
I will merge this PR and verify that issue will fix, if not I will revert that change.

benchmark_runner/common/ocp_resources/create_odf.py
RobertKrawitz: My concern is that this will work unreliably because it won't get at the root cause of the problem.

PR #691, opened 2023-11-05
Add sleep after disk deletion

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. Add sleep after Ceph disk deletion - for avoiding installation failure
2. ODF installation verification and raise an error if it fails


RobertKrawitz (approved):
Approving, but if possible, I'd like to find something more robust than a simple timeout.

PR #699, opened 2023-11-07
Fix grafana kata rpm

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Fix Grafana kata rpm version, parse the string into integer


benchmark_runner/grafana/perf/jsonnet/main.libsonnet
```diff
@@ -357,13 +357,13 @@ g.dashboard.new('PerfCI-Regression-Summary')
           + elasticsearch.withMetrics([
             elasticsearch.metrics.MetricAggregationWithSettings.Max.withField('ci_minutes_time')
             + elasticsearch.metrics.MetricAggregationWithSettings.Max.withId('1')
-            + elasticsearch.metrics.MetricAggregationWithSettings.Max.settings.withScript('Integer.parseInt(\"0\"+doc["kata_rpm_version.keyword"].value.replace(\".\",\"\").replace(\"r\",\"\").replace(\"c\",\"\").replace(\"f\",\"\").replace(\"-\",\"\").replace(\"e\",\"1\").replace(\"c\",\"\").replace(\"r\",\"\"))')
+            + elasticsearch.metrics.MetricAggregationWithSettings.Max.settings.withScript('(doc["kata_rpm_version.keyword"].value.indexOf(\" \") == -1) ? Integer.parseInt(\"0\"+doc["kata_rpm_version.keyword"].value.replace(\".\",\"\").replace(\"r\",\"\").replace(\"c\",\"\").replace(\"f\",\"\").replace(\"-\",\"\")) : 0')
```
RobertKrawitz, line 360: This really needs a unit test with as wide a variety as possible of versions and expected mappings.  Otherwise it's going to break very easily.

And I really think you need to find a way to store this as something other than an integer, since the name is intrinsically structured (how do you do it for the OCP version?).  I'm not going to hold this for that, though.

RobertKrawitz (changes requested):
Please add a unit test as noted.

benchmark_runner/grafana/perf/jsonnet/main.libsonnet
ebattat: Thats a grafonnet code, no unittest for it.

RobertKrawitz: We can test the parsing of the values, though.  It's just python code.

RobertKrawitz (changes requested):
Still would like a unit test to ensure that we can parse different values.

benchmark_runner/grafana/perf/jsonnet/main.libsonnet
ebattat: That should be in separate task, not part of this PR

RobertKrawitz: I disagree.  This is exactly the right point to add this unit test.  You've modified the parser, so there should be a test to ensure that the parser does the right thing.

ebattat: @RobertKrawitz, I disagree. 
I will not add unittest for 1 line.
Its a seperate project !!!

RobertKrawitz: I've had plenty of one liners that are so complex that a unit test is needed.  But OK, it isn't Python code, so it can be a separate issue.

RobertKrawitz (approved):
Approved; opened issue 702 for tracking unit test.

benchmark_runner/grafana/perf/jsonnet/main.libsonnet
ebattat: Yes, main.libsonnet is pure grafonnet code based on jsonnet.

PR #700, opened 2023-11-07
Revert "Work around virtiofsd threadpoolsize not working in 4.13"

RobertKrawitz:
Reverts redhat-performance/benchmark-runner#606

When OCP gets https://gitlab.com/virtio-fs/virtiofsd/-/merge_requests/174, this can be merged, but not until then.


ebattat (commented):
Do we need a special clusterbuster version to support it ?

RobertKrawitz:
> Do we need a special clusterbuster version to support it ?

No.

ebattat (approved):
/LGTM

PR #703, opened 2023-11-07
Add sleep between disk deletion

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Add sleep between disk deletion, look that not all the disk empty.


ebattat:
Seems not solving the issue

PR #704, opened 2023-11-08
Change operator install order

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Change operator install order due to ODF disk delete issue.
Install first 'lso' and 'odf' and then other operators.


RobertKrawitz (approved):
Understood that this is experimental.  It would be good to be able to test things like this without having to merge PRs (i. e. from a branch).

PR #706, opened 2023-11-08
Fix odf delete disk

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. Add method run_debug_node to run any command inside debug node
2. Use run_debug_node method to delete disks

### DIFF
--- a/benchmark_runner/common/oc/oc.py
+++ b/benchmark_runner/common/oc/oc.py
@@ -100,6 +100,16 @@ class OC(SSH):
         else:
             raise Exception('Missing free disk id')
 
+    @typechecked
+    def run_debug_node(self, node: str, cmd: str):
+        """
+        This method runs command in debug node
+        :param: node
+        :param: cmd
+        :return:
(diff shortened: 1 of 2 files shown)


benchmark_runner/common/oc/oc.py
```diff
@@ -100,6 +100,16 @@ def get_free_disk_id(self):
...
 
+    @typechecked
+    def run_debug_node(self, node: str, cmd: str):
+        """
+        This method runs command in debug node
+        :param: node
+        :param: cmd
+        :return:
+        """
+        self.run(f'{self.__cli} debug node/{node} -- chroot /host sh -c {cmd}')
```
RobertKrawitz, line 111: debug --no-tty=true

RobertKrawitz (changes requested):

RobertKrawitz (approved):

PR #707, opened 2023-11-09
Install last patch

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Installing only the last non approved patch to avoid installing incorrect patch as following:
```
$ oc get InstallPlan -n openshift-storage
NAME            CSV                          APPROVAL   APPROVED
install-1   mcg-operator.v4.12.9-1  Manual          false
install-2  ocs-operator.v4.13.4-2   Manual          false
```

### DIFF
--- a/benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
+++ b/benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
@@ -83,46 +83,46 @@ class CreateOCPResourceOperations:
             current_wait_time += OC.SLEEP_TIME
         raise OCPResourceCreationTimeout(resource)
 
-    def apply_non_approved_patch(self, approved_values_list: list, namespace: str, resource: str):
+    def apply_last_non_approved_patch(self, approved_values_list: list, namespace: str, resource: str):
         """
-        This method returns the index of not approved InstallPlan
+        This method applies last non approve patch
         :param approved_values_list:
         :param namespace:
         :param resource:
         :return:
         """
-        for index, approved in enumerate(approved_values_list):
-            # APPROVED false - need to patch
-            if approved == 'false':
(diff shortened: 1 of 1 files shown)


benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
```diff
@@ -83,46 +83,46 @@ def wait_for_ocp_resource_create(self, resource: str, verify_cmd: str, status: s
...
+            install_plan = self.__oc.run(cmd=install_plan)
+            install_plan_cmd = """ oc patch InstallPlan -n namespace install_plan -p '{"spec":{"approved":true}}' --type merge""".replace('install_plan', str(install_plan))
+            run_install_plan_cmd = install_plan_cmd.replace('namespace', namespace)
+            self.__oc.run(cmd=run_install_plan_cmd)
+            # verify current status
+            check_status = 'oc get InstallPlan -n namespace -ojsonpath={..spec.approved}'.replace('namespace', namespace).strip()
+            result = self.__oc.run(cmd=check_status).split()[last_index]
+            if result == 'true':
+                if resource == 'odf':
+                    for ind in range(3):
```
RobertKrawitz, line 108: Why `range(3)`?

Shouldn't you try to match the install plan to the OCP version?

ebattat: because in ODF it return 3 options

RobertKrawitz: Is it guaranteed to return exactly three options?

ebattat:
The issue is with Github actions

PR #715, opened 2023-11-11
Verify that CSVs in openshift-storage have succeeded

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Update number of ODF version status dynamically instead of hard coded number

### DIFF
--- a/benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
+++ b/benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
@@ -80,46 +80,33 @@ class CreateOCPResourceOperations:
             current_wait_time += OC.SLEEP_TIME
         raise OCPResourceCreationTimeout(resource)
 
-    def apply_non_approved_patch(self, approved_values_list: list, namespace: str, resource: str):
+    def apply_patch(self, namespace: str, resource: str):
         """
-        This method returns the index of not approved InstallPlan
-        :param approved_values_list:
+        This method applies an InstallPlan that has not been approved
         :param namespace:
         :param resource:
         :return:
         """
-        for index, approved in enumerate(approved_values_list):
+        install_plan_cmd = f"oc get InstallPlan -n {namespace} -ojsonpath={{$.items[*].metadata.name}}"
+        install_plan_names = self.__oc.run(cmd=install_plan_cmd).split()
+        for name in install_plan_names:
+            check_approved = f"oc get InstallPlan -n {namespace} {name} -ojsonpath={{..spec.approved}}"
+            approved = self.__oc.run(cmd=check_approved)
             # APPROVED false - need to patch
-            if approved == 'false':
-                install_plan = """ oc get InstallPlan -n namespace -ojsonpath={.items[index].metadata.name} """.replace('index', str(index))
-                install_plan = install_plan.replace('namespace', namespace)
-                install_plan = self.__oc.run(cmd=install_plan)
-                install_plan_cmd = """ oc patch InstallPlan -n namespace install_plan -p '{"spec":{"approved":true}}' --type merge""".replace('install_plan', str(install_plan))
-                run_install_plan_cmd = install_plan_cmd.replace('namespace', namespace)
-                self.__oc.run(cmd=run_install_plan_cmd)
+            if not approved:
+                install_plan_cmd = (f"oc patch InstallPlan -n {namespace} {name} -p '{{\"spec\":{{\"approved\":true}}}}' --type merge")
+                self.__oc.run(cmd=install_plan_cmd)
                 # verify current status
-                check_status = 'oc get InstallPlan -n namespace -ojsonpath={..spec.approved}'.replace('namespace', namespace).strip()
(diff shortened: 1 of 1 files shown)


benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
```diff
@@ -102,7 +102,8 @@ def apply_non_approved_patch(self, approved_values_list: list, namespace: str, r
                 result = self.__oc.run(cmd=check_status).split()[index]
                 if result == 'true':
                     if resource == 'odf':
-                        for ind in range(3):
+                        num_status = self.__oc.run("oc get csv -n openshift-storage -ojsonpath={$.items[*].status.phase}")
+                        for ind in range(len(num_status.split())):
```
RobertKrawitz, line 106: The intent is to ensure that all CSVs in openshift-storage have succeeded?  If so, please update the title accordingly.

But you really want metadata.name rather than status.phase at line 105

RobertKrawitz (changes requested):

benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
ebattat: We need the number of CSV status that should be checked, actually no difference between status and name both are working

RobertKrawitz: It would be better to use the name, and then

```
verify_cmd=f"oc get csv -n {namespace} {name} -ojsonpath='{{.items[ind].status.phase}}'"
```

Then if the order of the CSVs in the list changes for some reason you're still OK.

RobertKrawitz (changes requested):

benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
RobertKrawitz: You're relying on the order of the returned CSVs to be the same from run to run.  Suppose it changes?

ebattat: No

RobertKrawitz: When I say "from run to run", I mean suppose the order returned is random?  You're assuming that index 1, for example, is going to be the same value each time.

RobertKrawitz: No what?

ebattat: You right, fixed it.

benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
```diff
@@ -91,24 +91,24 @@ def apply_non_approved_patch(self, approved_values_list: list, namespace: str, r
         for index, approved in enumerate(approved_values_list):
             # APPROVED false - need to patch
             if approved == 'false':
-                install_plan = """ oc get InstallPlan -n namespace -ojsonpath={.items[index].metadata.name} """.replace('index', str(index))
-                install_plan = install_plan.replace('namespace', namespace)
+                install_plan = f"oc get InstallPlan -n {namespace} -ojsonpath={{.items[index].metadata.name}}"
```
RobertKrawitz, line 94: Do the same thing here -- rather than relying on list ordering, do it by name.

RobertKrawitz (changes requested):

benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
```diff
@@ -80,46 +80,34 @@ def wait_for_ocp_resource_create(self, resource: str, verify_cmd: str, status: s
...
             # APPROVED false - need to patch
-            if approved == 'false':
-                install_plan = """ oc get InstallPlan -n namespace -ojsonpath={.items[index].metadata.name} """.replace('index', str(index))
-                install_plan = install_plan.replace('namespace', namespace)
-                install_plan = self.__oc.run(cmd=install_plan)
-                install_plan_cmd = """ oc patch InstallPlan -n namespace install_plan -p '{"spec":{"approved":true}}' --type merge""".replace('install_plan', str(install_plan))
-                run_install_plan_cmd = install_plan_cmd.replace('namespace', namespace)
-                self.__oc.run(cmd=run_install_plan_cmd)
+            if not approved:
+                install_plan_cmd = """ oc patch InstallPlan -n namespace name -p '{"spec":{"approved":true}}' --type merge""".replace(
```
RobertKrawitz, line 97:
```
install_plan_cmd = f"""oc patch InstallPlan -n {namespace} {name} -p {{"spec":{{"approved":true}}}}' --type merge"""
```
(if the pod name contains the word "namespace" the wrong thing will happen).

RobertKrawitz (changes requested):

RobertKrawitz (approved):

PR #740, opened 2023-11-15
fix operator phase check

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. Update check from boolean to string because the command for phase check return string 'false'
2. Add verification that 'rook-ceph-tools' is creating.

### DIFF
--- a/benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
+++ b/benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
@@ -93,12 +93,12 @@ class CreateOCPResourceOperations:
             check_approved = f"oc get InstallPlan -n {namespace} {name} -ojsonpath={{..spec.approved}}"
             approved = self.__oc.run(cmd=check_approved)
             # APPROVED false - need to patch
-            if not approved:
+            if not {'true': True, 'false': False}.get(approved.lower(), False):
                 install_plan_cmd = (f"oc patch InstallPlan -n {namespace} {name} -p '{{\"spec\":{{\"approved\":true}}}}' --type merge")
                 self.__oc.run(cmd=install_plan_cmd)
                 # verify current status
                 result = self.__oc.run(cmd=check_approved)
-                if result == 'true':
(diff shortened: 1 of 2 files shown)


benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
```diff
@@ -93,7 +93,7 @@ def apply_patch(self, namespace: str, resource: str):
             check_approved = f"oc get InstallPlan -n {namespace} {name} -ojsonpath={{..spec.approved}}"
             approved = self.__oc.run(cmd=check_approved)
             # APPROVED false - need to patch
-            if not approved:
+            if 'false' == approved:
```
RobertKrawitz, line 96: It might be better to have __oc.run() do the conversion to boolean, at least as an option.  At the very least, use a case-insensitive comparison in case this changes again.

RobertKrawitz (changes requested):

benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
ebattat: __oc.run() just run the command, not handle the result.
Do you think we need to force shifting to boolean ?

benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
```diff
@@ -93,12 +93,12 @@ def apply_patch(self, namespace: str, resource: str):
             check_approved = f"oc get InstallPlan -n {namespace} {name} -ojsonpath={{..spec.approved}}"
             approved = self.__oc.run(cmd=check_approved)
             # APPROVED false - need to patch
-            if not approved:
+            if not {'true': True, 'false': False}.get(approved.lower(), False):
```
RobertKrawitz, line 96: A bit more elaborate than needed, definitely works.

RobertKrawitz (approved):

PR #745, opened 2023-11-18
Fix ODF installation validation

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

The issue is that ODF 4 CSVs are not starting at the same time. This causes an issue in the CSV verification mechanism because, in the best case, the verification is for one CSV name, and in the worst case, there are no CSV names, leading to a failure in the storage cluster installation


benchmark_runner/common/oc/oc.py
```diff
@@ -244,16 +244,17 @@ def wait_for_dv_status(self,
...
-        self.run(f"oc patch storagecluster ocs-storagecluster -n {namespace} --type json --patch '[{{ \"op\": \"replace\", \"path\": \"/spec/enableCephTools\", \"value\": true }}]'")
-        rook_ceph_tools_pod = self._get_pod_name(pod_name='rook-ceph-tools', namespace=namespace)
-        self.wait_for_pod_create(pod_name=rook_ceph_tools_pod, namespace=namespace, timeout=self.timeout)
-        result = self.run(f"oc -n openshift-storage rsh {rook_ceph_tools_pod} ceph health")
-        return 'HEALTH_OK' == result.strip()
+        self.run(
+            f"oc patch storagecluster ocs-storagecluster -n {namespace} --type json --patch '[{{ \"op\": \"replace\", \"path\": \"/spec/enableCephTools\", \"value\": true }}]'")
+        self.wait_for_patch(pod_name='rook-ceph-tools', label='app=rook-ceph-tools', label_uuid=False, namespace=namespace)
+        health_check = self.run(
+            f"oc -n openshift-storage rsh {self._get_pod_name(pod_name='rook-ceph-tools', namespace=namespace)} ceph health")
```
RobertKrawitz, line 256: -n {{ namespace }}

benchmark_runner/common/oc/oc.py
```diff
@@ -648,6 +649,27 @@ def wait_for_initialized(self, label: str, workload: str = '', status: str = 'In
...
+        @param timeout:
+        @return:
+        """
+        current_wait_time = 0
+        while timeout <= 0 or current_wait_time <= timeout:
+            if self._get_pod_name(pod_name=pod_name, namespace=namespace) and self.wait_for_ready(label=label, label_uuid=label_uuid, namespace=namespace):
+                return True
+            # sleep for x seconds
+            time.sleep(OC.SLEEP_TIME)
+            current_wait_time += OC.SLEEP_TIME
```
RobertKrawitz, line 670: Rather than rely on OC.SLEEP_TIME here, use time.time() to get the actual time.

benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
```diff
@@ -12,6 +12,9 @@ class CreateOCPResourceOperations:
     """
     This class creates OCP resources
     """
+    # Number of CSVs that should be validated
+    ODF_CSV_NUM = 4
```
RobertKrawitz, line 16: You have the CSV names, correct?  Check them by name (make sure that all of the expected CSVs are validated) rather than count.

benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
```diff
@@ -53,33 +56,52 @@ def wait_for_ocp_resource_create(self, resource: str, verify_cmd: str, status: s
...
             current_wait_time += OC.SLEEP_TIME
         raise OCPResourceCreationTimeout(resource)
 
+    def verify_csv_installation(self, namespace: str, resource: str):
+        """
+        This method verifies csv installation
+        :param namespace:
+        :param resource:
+        """
+        csv_names = self.__oc.run(f"oc get csv -n {namespace} -ojsonpath={{$.items[*].metadata.name}}")
```
RobertKrawitz, line 99: `'-ojsonpath={{$.items[*].metadata.name}}'`

so no problems with metacharacter expansion.

RobertKrawitz (changes requested):

benchmark_runner/common/oc/oc.py

benchmark_runner/common/oc/oc.py
ebattat: We use sleep of 3 sec between each operation waiting that the patch pod will be started

benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
ebattat: I prefer to keep it generic as we can to avoid failures
We just need to install ODF and not to test it.

benchmark_runner/common/oc/oc.py
RobertKrawitz: This isn't that big of a deal (if your timeout runs long, it's not really a problem), but don't assume that each cycle will take exactly 3 seconds.

ebattat: Its for short period tasks, like create pod.

benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
ebattat: No already test it and it works

benchmark_runner/common/ocp_resources/create_ocp_resource_operations.py
ebattat: @RobertKrawitz, I have fixed.

RobertKrawitz (approved):

PR #752, opened 2023-12-05
Update oc to cli

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Update oc to cli

### DIFF
--- a/benchmark_runner/common/oc/oc.py
+++ b/benchmark_runner/common/oc/oc.py
@@ -53,7 +53,7 @@ class OC(SSH):
         This method returns cnv version
         :return:
         """
-        return self.run(f"{self.__cli} get csv -n openshift-cnv $(oc get csv -n openshift-cnv --no-headers | awk '{{ print $1; }}') -ojsonpath='{{.spec.version}}'")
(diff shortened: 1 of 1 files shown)


RobertKrawitz (changes requested):
Good!  There are some others, though:
- create_kata.py (around lines 39 and 56)
- create_ocp_resource_operations.py (around lines 97 and 100)

ebattat:
@RobertKrawitz,

This change is relevant only OC class and not to Operator class because the operators are relevant only to OCP and not k8s, I just change all the oc methods to use the same approach.

RobertKrawitz:
Those others all use the OC class though; you might as well fix it all while you're at it.

ebattat:
There is no need to change all other 'oc' commands because 'csv,' 'kata,' and 'odf' are only relevant to OpenShift. This means there is no need to use 'kubectl' commands instead of 'oc'.

RobertKrawitz:
> @RobertKrawitz, We don't need it for OpenShift operators; we only use the 'oc' command for them

Yeah, OK, I'd prefer to keep it general, but I see the point.

RobertKrawitz (approved):

PR #755, opened 2024-01-07
Fix bootstorm workload

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. Reduce Windows vms from 120 to 111 due to Insufficient memory issue. => cause to p90 degradation.
2. Remove node selector from windows and fedora yamls for better performance results.
3. Add total run time metric - calculate the total run time metric from the first running vm till the last vm
4. Update Grafana panel with total run time metric.


bbenshab (approved):
LGTM

PR #756, opened 2024-01-08
VM memory distribution prometheus metrics

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Add Prometheus VM memory metrics capturing:

- virt-handler
- virt-operator 
- virt-controller
- virt-api

Also adding exceeds requested memory for each metric.

### DIFF
--- a/benchmark_runner/common/prometheus/metrics-default.yaml
+++ b/benchmark_runner/common/prometheus/metrics-default.yaml
@@ -1,5 +1,3 @@
-
-
 # container CPU/Memory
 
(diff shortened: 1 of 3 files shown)


bbenshab:
I don't think we need to show any of those metrics in the Grafana, it's just TMI , however, we should save that data in the artifact in a case that we might need it for debugging.

ebattat:
@bbenshab, Agree that the reason I didnt add it to the Grafana dashboard

RobertKrawitz (approved):

jeniferh (approved):
LGTM

PR #758, opened 2024-01-14
Fix summary report

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. Add hammerdb LSO run result (PostgreSQL)
2. Add storage type column
3. Move dashboard url to bottom
4. Add dates/uuid in version table
5. colorized scale for percentages accordingly: [positive: green]
```
            if val < -50:
                color = 'red'
            elif val < -10:
                color = 'orange'
            elif val < 0:
                color = 'black'
            else:
                color = 'green'
```
7. Update html report with relevant versions - plan to upload it in the future to shared storage

### DIFF
--- a/benchmark_runner/jupyterlab/templates/summary_report/summary_report.ipynb
+++ b/benchmark_runner/jupyterlab/templates/summary_report/summary_report.ipynb
@@ -11,8 +11,8 @@
    ]
   },
   {
-   "cell_type": "raw",
-   "id": "14a344ae-ba02-4690-9656-ebb13db1d5be",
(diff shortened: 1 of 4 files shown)


RobertKrawitz (approved):
I see nothing wrong with this, but I suggest having someone more familiar with this area review it also.

jeniferh (approved):
LGTM, I can't provide much input on the .ipynb changes themselves but the summary report you shared looks great

PR #763, opened 2024-01-21
Save scale logs error

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

When a scale run fails, no logs are uploaded to S3. After a thorough investigation, it was discovered that when running methods in multiple threads, exceptions must be handled inside the method and not outside. That's the reason separate exceptions were added in every run scale method


RobertKrawitz (approved):

ebattat:
@RobertKrawitz, I have enclosed all the methods running in parallel within try and except blocks.

RobertKrawitz:
Yes, understood, the PR writeup was clear and I approved it yesterday.

PR #764, opened 2024-01-22
Fix vdbench pod failure

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

The Vdbench pod workload sometimes fails due to a float casting error:
```
ile "/usr/local/lib/python3.10/site-packages/benchmark_runner/workloads/workloads_operations.py", line 219, in __is_float
    float(value)
TypeError: float() argument must be a string or a real number, not 'list'
```
Adding list check to __is_float method to avoid the above error.

For more error details see [logs](https://github.com/redhat-performance/benchmark-runner/actions/runs/7613842210/job/20737537991)

### DIFF
--- a/benchmark_runner/workloads/workloads_operations.py
+++ b/benchmark_runner/workloads/workloads_operations.py
@@ -212,13 +212,13 @@ class WorkloadsOperations:
     def __is_float(value) -> bool:
         """
         This method checks if value is float
-        :param value:
-        :return:
+        :param value: The value to check
+        :return: True if the value is a float, False when cannot be converted to a float
         """
         try:
             float(value)
             return True
-        except ValueError:
+        except (ValueError, TypeError):
             return False
 
     def _create_scale_logs(self):


benchmark_runner/workloads/workloads_operations.py
```diff
@@ -211,15 +211,18 @@ def _get_run_artifacts_hierarchy(self, workload_name: str = '', is_file: bool =
...
     def __is_float(value) -> bool:
         """
-        This method checks if value is float
-        :param value:
-        :return:
+        This method checks if value is float or a list of floats
+        :param value: The value to check
+        :return: True if the value is a float, False if it's a list or cannot be converted to a float
         """
-        try:
```
RobertKrawitz, line 218: This function can be greatly simplified:
```
try:
    float(value)
    return True
except (TypeError, ValueError):
    return False

RobertKrawitz (changes requested):

RobertKrawitz (approved):

PR #766, opened 2024-02-02
Update GTA plugins

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [ ] enhancement
- [ ] documentation
- [x] dependencies

## Description

Update the GHA plugins to use the node-20, for that we need to update the current plugin versions.

ref:

[GH checkout](https://github.com/marketplace/actions/checkout)
[Python plugin](https://github.com/actions/setup-python)

Solved plugins warnings:
[Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/setup-python@v4. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.](https://github.com/redhat-performance/benchmark-runner/actions/runs/7752236376)

### DIFF
--- a/.github/workflows/Build_runner.yml
+++ b/.github/workflows/Build_runner.yml
@@ -25,9 +25,9 @@ jobs:
     name: pypi_upload
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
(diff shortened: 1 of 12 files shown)


athiruma (approved):
/lgtm

PR #768, opened 2024-02-08
PerfLab OpenShift and Operators deployment

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. Add OCP installation using Jetlag BYOL installation
2. Add OCP operators installation LSO, ODF and CNV Nightly
3. Add bare metal class to handle all operation and exceptions for On-premise and IBM Cloud servers
4. Add Jenkinsfile for Openshift and Operators deployment
5. Add disk maxSize: 1250Ti for choosing only SSD disks in IBM cloud or in PerfLab

TO-DO after benchmark-runner packaging: 
- Need to verify that 2 Jenkins jobs (OCP/Operators deployment) are working using benchmark-runner container after uploading to quay.io
- Need to verify that no changes in IBM clouds installation

### DIFF
--- a/benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
+++ b/benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
@@ -1,0 +1,300 @@
+
+import time
+from datetime import datetime
+from typeguard import typechecked
+import ast  # change string list to list
+
+from benchmark_runner.common.logger.logger_time_stamp import logger_time_stamp, logger, datetime_format
+from benchmark_runner.main.environment_variables import environment_variables
+from benchmark_runner.common.remote_ssh.remote_ssh import ConnectionData, RemoteSsh
+from benchmark_runner.common.ocp_resources.create_ocp_resource import CreateOCPResource
+from benchmark_runner.common.oc.oc import OC
+from benchmark_runner.common.github.github_operations import GitHubOperations
+from benchmark_runner.common.clouds.BareMetal.bare_metal_exceptions import MissingMasterNodes, MissingWorkerNodes, OCPInstallationFailed
+from benchmark_runner.common.ssh.ssh import SSH
+from benchmark_runner.common.assisted_installer.assisted_installer_latest_version import AssistedInstallerVersions
+
+
+class BareMetalOperations:
+    """
+    This class is responsible for Bare metal OpenShift installation operations, all commands run on remote provision server
+    """
+    LATEST = 'latest'
+    SHORT_TIMEOUT = 600
+
+    @typechecked
+    def __init__(self, user: str):
+        self._environment_variables_dict = environment_variables.environment_variables_dict
+        self._user = user
+        self._ocp_env_flavor = self._environment_variables_dict.get('ocp_env_flavor', '')
+        self._create_pod_ci_cmd = self._environment_variables_dict.get('create_pod_ci_cmd', '')
+        self._provision_kubeadmin_password_path = self._environment_variables_dict.get('provision_kubeadmin_password_path', '')
+        self._provision_kubeconfig_path = self._environment_variables_dict.get('provision_kubeconfig_path', '')
+        self._provision_installer_path = self._environment_variables_dict.get('provision_installer_path', '')
+        self._provision_installer_cmd = self._environment_variables_dict.get('provision_installer_cmd', '')
+        self._provision_installer_log = self._environment_variables_dict.get('provision_installer_log', '')
+        self._install_ocp_version = self._environment_variables_dict.get('install_ocp_version', '')
+        self._cluster_type = self._environment_variables_dict.get('cluster_type', '')
+        self._expected_nodes = self._environment_variables_dict.get('expected_nodes', '')
+        if self._expected_nodes:
+            # string to dict
+            self._expected_nodes = ast.literal_eval(self._expected_nodes)
+        self._ocp_version_build = self._environment_variables_dict.get('ocp_version_build', '')
+        self._num_odf_disks = int(self._environment_variables_dict.get('num_odf_disk', 1))
+        self._provision_ip = self._environment_variables_dict.get('provision_ip', '')
+        self._provision_port = int(self._environment_variables_dict.get('provision_port', ''))
+        self._provision_timeout = int(self._environment_variables_dict.get('provision_timeout', ''))
+        self._container_private_key_path = self._environment_variables_dict.get('container_private_key_path', '')
+        self._connection_data = ConnectionData(host_name=self._provision_ip,
+                                                user_name=self._user,
+                                                port=self._provision_port,
+                                                timeout=self._provision_timeout,
+                                                ssh_key=self._container_private_key_path)
+        self._remote_ssh = RemoteSsh(self._connection_data)
+        if self._environment_variables_dict.get('github_repository_short', ''):
+            self._github_operations = GitHubOperations()
+        self._ssh = SSH()
+        self._cli = self._environment_variables_dict.get('cli', '')
+
+    def _get_kubeadmin_password(self):
+        """
+        This method returns kubeadmin password if exist
+        :return:
+        """
+        if self._remote_ssh.exist(remote_path=self._provision_kubeadmin_password_path):
+            return self._remote_ssh.run_command(f'cat {self._provision_kubeadmin_password_path}')
+
+    def _get_kubeconfig(self):
+        """
+        This method returns kubeconfig if exist
+        :return:
+        """
+        if self._remote_ssh.exist(remote_path=self._provision_kubeconfig_path):
+            return self._remote_ssh.run_command(f'cat {self._provision_kubeconfig_path}')
+
+    def _wait_for_install_complete(self, sleep_time: int = SHORT_TIMEOUT):
+        """
+        This method waits till ocp install complete
+        :param sleep_time:
+        :return:
+        """
+        current_wait_time = 0
+        install_log = ''
+        while self._provision_timeout <= 0 or current_wait_time <= self._provision_timeout:
+            install_log = self._remote_ssh.run_command(self._provision_installer_log)
+            if 'failed=0' in install_log:
+                return True
+            elif 'failed=1' in install_log:
+                return False
+            logger.info(f'Waiting till OCP install complete, waiting {int(current_wait_time/60)} minutes')
+            # sleep for x seconds
+            time.sleep(sleep_time)
+            current_wait_time += sleep_time
+        raise OCPInstallationFailed(install_log)
+
+    def _install_ocp_cmd(self):
+        """
+        This method returns local ocp installer cmd if exist
+        :return:
+        """
+        if self._remote_ssh.exist(remote_path=self._provision_installer_path):
+            return self._provision_installer_cmd
+
+    @logger_time_stamp
+    def connect_to_provisioner(self):
+        """
+        This method connect to remote provision machine
+        :return:
+        """
+        self._remote_ssh.connect()
+
+    @logger_time_stamp
+    def disconnect_from_provisioner(self):
+        """
+        This method
+        :return:
+        """
+        self._remote_ssh.disconnect()
+
+    def _get_installation_version(self):
+        """
+        This method return installation version
+        :return:
+        """
+        assisted_installer_versions = AssistedInstallerVersions()
+        if self.LATEST in self._install_ocp_version:
+            openshift_version_data = self._install_ocp_version.split('-')
+            # release candidate or engineering candidate or feature candidate version
+            if '-rc' in self._install_ocp_version or '-ec' in self._install_ocp_version or '-fc' in self._install_ocp_version:
+                return assisted_installer_versions.get_latest_version(latest_version=f'{openshift_version_data[1]}-{openshift_version_data[2]}')
+            # release version
+            else:
+                return assisted_installer_versions.get_latest_version(latest_version=openshift_version_data[1])
+        else:
+            return self._install_ocp_version
+
+    def get_ocp_server_version(self):
+        """
+        This method return ocp server version
+        :return:
+        """
+        return self._remote_ssh.run_command(command=f"{self._cli} version -ojson | jq -r '.openshiftVersion'").strip()
+
+    def version_already_installed(self):
+        """
+        This method validate if version is already install
+        :return: True if it already installed, False if it NOT already installed
+        """
+        if self.get_ocp_server_version() == 'null':
+            return False
+        elif self.LATEST in self._install_ocp_version and self.get_ocp_server_version() == self._get_installation_version():
+            return self.get_ocp_server_version().strip()
+        elif self._install_ocp_version == self.get_ocp_server_version():
+            return self._install_ocp_version
+        else:
+            return False
+
+    @logger_time_stamp
+    def restart_pod_ci(self):
+        """
+        This method restart pod ci: elastic, kibana, grafana, nginx and flask - solved connectivity issue after installation
+        :return:
+        """
+        self._remote_ssh.run_command('podman pod restart pod_ci')
+
+    @typechecked()
+    @logger_time_stamp
+    def update_ocp_version(self, file_name: str):
+        """
+        This method update the ocp version on provision machine
+        :param file_name:
+        :return:
+        """
+        # Get the latest assisted installer version
+        if self.LATEST in self._install_ocp_version:
+            self._install_ocp_version = self._get_installation_version()
+        openshift_version_data = self._install_ocp_version.split('.')
+        self._remote_ssh.replace_parameter(remote_path='/root/jetlag/ansible/vars',
+                                            file_name=file_name,
+                                            parameter='ocp_release_image:',
+                                            value=f'quay.io\/openshift-release-dev\/ocp-release:{self._install_ocp_version}-x86_64')
+        self._remote_ssh.replace_parameter(remote_path='/root/jetlag/ansible/vars',
+                                            file_name=file_name,
+                                            parameter='openshift_version:',
+                                            value=f'"{openshift_version_data[0]}.{openshift_version_data[1]}"')
+
+    def update_provision_config(self):
+        """
+        This method update provision config file
+        :return:
+        """
+        # update ssh config - must add it and also mount it
+        self._ssh.run(f"echo Host provision > /{self._user}/.ssh/config")
+        self._ssh.run(f"echo -e '\t'HostName {self._provision_ip} >> /{self._user}/.ssh/config")
+        self._ssh.run(f"echo -e '\t'User {self._user} >> /{self._user}/.ssh/config")
+        self._ssh.run(f"echo -e '\t'IdentityFile {self._container_private_key_path} >> /{self._user}/.ssh/config")
+        self._ssh.run(f"echo -e '\t'StrictHostKeyChecking no >> /{self._user}/.ssh/config")
+        self._ssh.run(f"echo -e '\t'ServerAliveInterval 30 >> /{self._user}/.ssh/config")
+        self._ssh.run(f"echo -e '\t'ServerAliveCountMax 5 >> /{self._user}/.ssh/config")
+        self._ssh.run(f"chmod 600 /{self._user}/.ssh/config")
+
+    @logger_time_stamp
+    def run_ocp_installer(self):
+        """
+        This method run ocp assisted installer
+        :return: True if installation success and raise exception if installation failed
+        """
+        """
+        This method run ocp assisted installer
+        :return: True if installation success and raise exception if installation failed
+        """
+        self.update_provision_config()
+        logger.info(f'Starting OCP assisted installer, Start time: {datetime.now().strftime(datetime_format)}')
+        # Must add -t otherwise remote ssh of ansible will not end
+        self._ssh.run(f"ssh -t provision \"{self._install_ocp_cmd()}\" ")
+        logger.info(f'OpenShift cluster {self._get_installation_version()} version is installed successfully, End time: {datetime.now().strftime(datetime_format)}')
+
+    @logger_time_stamp
+    def verify_install_complete(self):
+        """
+        This method verify that installation complete
+        :return: True if installation success and raise exception if installation failed
+        """
+        complete = self._wait_for_install_complete()
+        if not complete:
+            install_log = self._remote_ssh.run_command(self._provision_installer_log)
+            raise OCPInstallationFailed(install_log)
+        else:
+            return True
+
+    @logger_time_stamp
+    def oc_login(self):
+        """
+        This method login to the cluster with new credentials
+        :return:
+        """
+        oc = OC(kubeadmin_password=self._get_kubeadmin_password())
+        oc.login()
+        return oc
+
+    @staticmethod
+    def get_missing_node(expected_nodes: list, actual_nodes: list):
+        """
+        This method returns missing nodes
+        :return:
+        """
+        return set(expected_nodes) - set(actual_nodes)
+
+    @logger_time_stamp
+    def verify_cluster_is_up(self, oc: OC):
+        """
+        This method verify that master/worker nodes are up and running
+        :return:
+        """
+        master_nodes = oc.get_master_nodes()
+        if len(master_nodes.split()) == len(self._expected_nodes['master']):
+            logger.info('master nodes are up and running')
+            if self._ocp_env_flavor == 'PERF' and not self._cluster_type == 'SNO':
+                worker_nodes = oc.get_worker_nodes()
+                # if there are worker nodes, they must be 3
+                if len(worker_nodes.split()) > 1:
+                    if len(worker_nodes.split()) == len(self._expected_nodes['worker']):
+                        logger.info('worker nodes are up and running')
+                    else:
+                        missing_node = self.get_missing_node(expected_nodes=self._expected_nodes['worker'], actual_nodes=worker_nodes.split())
+                        raise MissingWorkerNodes(missing_node)
+        else:
+            missing_node = self.get_missing_node(expected_nodes=self._expected_nodes['master'], actual_nodes=master_nodes.split())
+            raise MissingMasterNodes(missing_node)
+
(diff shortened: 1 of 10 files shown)


benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
```diff
@@ -0,0 +1,32 @@
+
+class BMError(Exception):
```
RobertKrawitz, line 2: Name everything BareMetal rather than BM.

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
```diff
@@ -0,0 +1,32 @@
...
+        All exceptions raised by the benchmark runner library should inherit from this class. """
+    pass
+
+
+class MissingMasterNodes(BMError):
+    """
+    This class exception is for missing master nodes
+    """
+    def __init__(self):
+        self.message = 'There are missing master nodes'
```
RobertKrawitz, line 13: If you can identify which nodes are missing, enumerate them here.

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
```diff
@@ -0,0 +1,32 @@
...
+        super(MissingMasterNodes, self).__init__(self.message)
+
+
+class MissingWorkerNodes(BMError):
+    """
+    This class exception is for missing worker nodes
+    """
+    def __init__(self):
+        self.message = 'There are missing worker nodes'
```
RobertKrawitz, line 22: Likewise.

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
```diff
@@ -0,0 +1,32 @@
...
+        super(MissingWorkerNodes, self).__init__(self.message)
+
+
+class OCPInstallationFailed(BMError):
+    """
+    This class exception is notice that ibm ocp installation failed
+    """
+    def __init__(self):
+        self.message = f'Bare Metal OCP installation failed'
```
RobertKrawitz, line 31: If you can provide a more specific message about what failed, please do so.

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
```diff
@@ -0,0 +1,32 @@
...
+        super(OCPInstallationFailed, self).__init__(self.message)
```
RobertKrawitz, line 32: Need a newline to terminate.

benchmark_runner/common/clouds/BM/bare_metal_operations.py
```diff
@@ -0,0 +1,284 @@
...
+                return assisted_installer_versions.get_latest_version(latest_version=openshift_version_data[1])
+        else:
+            return self._install_ocp_version
+
+    def get_ocp_server_version(self):
+        """
+        This method return ocp server version
+        :return:
+        """
+        return self._remote_ssh.run_command(command=f"{self._cli} version -ojson | jq -r '.openshiftVersion'").strip()
```
RobertKrawitz, line 135: Unfortunate that this doesn't take -ojsonpath, but nothing you can do about it.  Might want to add a comment that `-ojsonpath` doesn't work for `oc version` so nobody tries to change it later.

benchmark_runner/common/clouds/BM/bare_metal_operations.py
```diff
@@ -0,0 +1,284 @@
...
+        :return:
+        """
+        # update ssh config - must add it and also mount it
+        self._ssh.run(f"echo Host provision > /{self._user}/.ssh/config")
+        self._ssh.run(f"echo -e '\t'HostName {self._provision_ip} >> /{self._user}/.ssh/config")
+        self._ssh.run(f"echo -e '\t'User {self._user} >> /{self._user}/.ssh/config")
+        self._ssh.run(f"echo -e '\t'IdentityFile {self._container_private_key_path} >> /{self._user}/.ssh/config")
+        self._ssh.run(f"echo -e '\t'StrictHostKeyChecking no >> /{self._user}/.ssh/config")
+        self._ssh.run(f"echo -e '\t'ServerAliveInterval 30 >> /{self._user}/.ssh/config")
+        self._ssh.run(f"echo -e '\t'ServerAliveCountMax 5 >> /{self._user}/.ssh/config")
```
RobertKrawitz, line 192: Better if you can construct the file here and then scp/cat it onto the remote file.

benchmark_runner/common/clouds/BM/bare_metal_operations.py
```diff
@@ -0,0 +1,284 @@
...
+        self._ssh.run(f"chmod 600 /{self._user}/.ssh/config")
```
RobertKrawitz, line 193: Ideally you'd `umask 77` beforehand, but there isn't really anything secret in this file.

benchmark_runner/common/clouds/BM/bare_metal_operations.py
```diff
@@ -0,0 +1,284 @@
...
+        :return: True if installation success and raise exception if installation failed
+        """
+        """
+        This method run ocp assisted installer
+        :return: True if installation success and raise exception if installation failed
+        """
+        self.update_provision_config()
+        logger.info(f'Starting OCP assisted installer, Start time: {datetime.now().strftime(datetime_format)}')
+        # Must add -t otherwise remote ssh of ansible will not end
+        self._ssh.run(f"ssh -t provision \"{self._install_ocp_cmd()}\" ")
```
RobertKrawitz, line 208: Does this raise an exception if it fails?  If so, you should catch the exception here.  You can re-raise it, but if the ocp installer fails, it's cleaner to catch it here.

benchmark_runner/common/ocp_resources/create_odf.py
```diff
@@ -34,9 +34,9 @@ def create_odf(self):
                     for node, disk_ids in self.__worker_disk_ids.items():
                         for disk_id in disk_ids:
                             disk = f'/dev/disk/by-id/{self.__worker_disk_prefix}{disk_id}'
-                            delete_node_disk += f"sgdisk --zap-all {disk}; wipefs -a {disk}; dd if=/dev/zero of='{disk}' bs=1M count=100 oflag=direct,dsync; blkdiscard {disk}; partprobe {disk};"
-                        self.__oc.run(cmd=f'chmod +x {os.path.join(self.__path, resource)}; {self.__path}/./{resource} "{node}" "{delete_node_disk}"')
-                        delete_node_disk = ''
+                            delete_node_disk += f"""sudo sgdisk --zap-all {disk}; sudo wipefs -a {disk}; sudo dd if=/dev/zero of='{disk}' bs=1M count=100 oflag=direct,dsync; sudo blkdiscard {disk}; sudo partprobe {disk};"""
```
RobertKrawitz, line 37: How do you ensure that sudo won't ask for a password here?

benchmark_runner/common/ocp_resources/odf/template/04_local_volume_set_template.yaml
```diff
@@ -19,7 +19,7 @@ spec:
     deviceTypes:
       - disk
       - part   # <-- Remove this if not using partitions
-    #deviceMechanicalProperties:
+    deviceMechanicalProperties:
     #  - NonRotational   # <-- For HDD change to Rotational
     #minSize: 0Ti   # <-- Uncomment and modify to limit the minimum size of disk used
-    #maxSize: 0Ti   # <-- Uncomment and modify to limit the maximum size of disk used
+    maxSize: 1250Ti   # <-- Uncomment and modify to limit the maximum size of disk used
```
RobertKrawitz, line 25: This isn't a very clean way to distinguish between different types of disks.

benchmark_runner/main/main.py
```diff
@@ -176,23 +202,23 @@ def run_clusterbuster_workload():
...
-        else:
-            logger.info(f'{ibm_operations.get_ocp_server_version().strip()} version already installed, skip OCP installation')
+        if install_step == 'run_bm_ocp_installer':
+            install_ocp_bare_metal(step=install_step)
+        elif install_step == 'run_ibm_ocp_installer':
+            install_ocp_ibm_cloud(step=install_step)
+        elif install_step == 'verify_bm_install_complete':
+            install_ocp_bare_metal(step=install_step)
+        elif install_step == 'verify_ibm_install_complete':
+            install_ocp_ibm_cloud(step=install_step)
```
RobertKrawitz, line 221: Need to handle invalid install step values.

jenkins/PerfCI_OpenShift_Deployment/Jenkinsfile
```diff
@@ -0,0 +1,114 @@
...
+        CONFIG_PATH = '/home/jenkins/.ssh/config'
+        CONTAINER_PRIVATE_KEY_PATH = '/root/.ssh/provision_private_key'
+        CONTAINER_CONFIG_PATH = '/root/.ssh/config'
+        PROVISION_TIMEOUT = 7200
+    }
+    stages {
+        stage('Cleanup') {
+            steps {
+                script {
+                    sh 'sudo podman ps -a --filter "name=benchmark-runner" -q | xargs -r podman rm'
```
RobertKrawitz, line 29: That won't run `xargs -r podman rm` privileged; the `sudo` will only apply to the pipeline component it references.

jenkins/PerfCI_OpenShift_Deployment/Jenkinsfile
```diff
@@ -0,0 +1,114 @@
...
+                }
+            }
+        }
+
+        stage('⚙️ SET SSH key') {
+            steps {
+                script {
+                    sh 'mkdir -p $WORKSPACE/.ssh/'
+                    withCredentials([file(credentialsId: 'perfci_provision_private_key_file', variable: 'PROVISION_PRIVATE_KEY_FILE')]) {
+                        sh "sudo cp \$PROVISION_PRIVATE_KEY_FILE $WORKSPACE/.ssh/provision_private_key"
```
RobertKrawitz, line 39: *This* needs to run under `umask 77` rather than a later chmod.  This is a private value that you do not want even momentarily exposed.

jenkins/PerfCI_Operators_Deployment/Jenkinsfile
```diff
@@ -0,0 +1,138 @@
...
+        KUBEADMIN_PASSWORD = readFile('/home/jenkins/.kube/kubeadmin-password').trim()
+        KUBEADMIN_PASSWORD_PATH = '/home/jenkins/.kube/kubeadmin-password'
+        KUBECONFIG_PATH = '/root/.kube/config'
+        PRIVATE_KEY_PATH = '/home/jenkins/.ssh/provision_private_key'
+        CONFIG_PATH = '/home/jenkins/.ssh/config'
+        WORKSPACE='/home/jenkins'
+        CONTAINER_PRIVATE_KEY_PATH = '/root/.ssh/provision_private_key'
+        PROVISION_PORT = 22
+        CNV_VERSION = '4.14'
+        ODF_VERSION = '4.13'
```
RobertKrawitz, line 20: These should go into the environment variable or secrets list; versions should not be scattered about the code.

RobertKrawitz (changes requested):

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
ebattat: Done

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
ebattat: Its not so important which master node is missing, cluster is corrupted and the user should investigate it by himself

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
ebattat: Its not so important which master node is missing, cluster is corrupted and the user should investigate it by himself

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
ebattat: Fixed

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
ebattat: Fixed

jenkins/PerfCI_Operators_Deployment/Jenkinsfile
ebattat: Thats the way we are doing it also in IBM Clouds, the versions are visible:
https://github.com/redhat-performance/benchmark-runner/blob/5781a5d67a14c3de0030ff923709e47f584d0148/.github/workflows/Weekly_Perf_Env_Operator_CI.yml#L80-L81

I believe it's unnecessary to keep the version a secret; knowing which version we have should not be a concern.

jenkins/PerfCI_OpenShift_Deployment/Jenkinsfile
ebattat: Got the following error without sudo: 
`ERRO[0000] XDG_RUNTIME_DIR directory "/run/user/0" is not owned by the current user`

jenkins/PerfCI_OpenShift_Deployment/Jenkinsfile
ebattat: Must add sudo because Jenkins script run with jenkins user,
Got the following error without sudo: 
`ERRO[0000] XDG_RUNTIME_DIR directory "/run/user/0" is not owned by the current user`

benchmark_runner/main/main.py
ebattat: Fixed

benchmark_runner/common/ocp_resources/odf/template/04_local_volume_set_template.yaml
ebattat: Right, I will modify it to filter by vendor and model:
    models:
      - PERC H730P Adp
    vendors:
      - DELL

ebattat:
@RobertKrawitz, pls let me know if you have more comments

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
RobertKrawitz: Please list which node(s) are missing.  It could be a transitory or infrequent error, where having that information may help in debugging.

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
RobertKrawitz: Same, please list what node(s) are missing.

benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
```diff
@@ -0,0 +1,284 @@
...
+    def _install_ocp_cmd(self):
+        """
+        This method returns local ocp installer cmd if exist
+        :return:
+        """
+        if self._remote_ssh.exist(remote_path=self._provision_installer_path):
+            return self._provision_installer_cmd
+
+    @logger_time_stamp
+    def connect(self):
```
RobertKrawitz, line 98: Name this `connect_to_provisioner` or similar.

benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
```diff
@@ -0,0 +1,284 @@
...
+        """
+        This method connect to remote provision machine
+        :return:
+        """
+        self._remote_ssh.connect()
+
+    @logger_time_stamp
+    def disconnect(self):
```
RobertKrawitz, line 106: Name this `disconnect_from_provisioner` or similar.

benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
```diff
@@ -0,0 +1,284 @@
...
+
+    @logger_time_stamp
+    def verify_install_complete(self):
+        """
+        This method verify that installation complete
+        :return: True if installation success and raise exception if installation failed
+        """
+        complete = self._wait_for_install_complete()
+        if not complete:
+            raise Exception(f'Installation failed')
```
RobertKrawitz, line 219: Can you get any information from _wait_from_install_complete (such as messages from the installer)?  If so, please print at least the last 10 lines or so of that output.

benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
```diff
@@ -0,0 +1,284 @@
...
+
+    @logger_time_stamp
+    def verify_cluster_is_up(self, oc: OC):
+        """
+        This method verify that master/worker nodes are up and running
+        :return:
+        """
+        master_nodes = oc.get_master_nodes()
+        # master always are 3
+        if len(master_nodes.split()) == 3:
```
RobertKrawitz, line 241: If you install on an SNO cluster, you'll only have 1.

benchmark_runner/main/main.py
```diff
@@ -96,36 +100,58 @@ def install_ocp(ibm_operations: IBMOperations, step: str):
...
     # convert str to list
     try:
         resources = ast.literal_eval(install_resources_list) if isinstance(install_resources_list, str) else install_resources_list
     except (ValueError, SyntaxError):
         resources = [install_resources_list]
-    logger.info(f'Start IBM OCP resources installation')
-    oc = ibm_operations.oc_login()
-    ibm_operations.verify_cluster_is_up(oc)
+    logger.info(f'Start BM OCP resources installation')
+    oc = bm_operations.oc_login()
```
RobertKrawitz, line 143: "bare_metal"

jenkins/PerfCI_OpenShift_Deployment/Jenkinsfile
RobertKrawitz: I understand you need to use sudo here; I think you probably also need to use `sudo xargs -r podman rm`

jenkins/PerfCI_OpenShift_Deployment/Jenkinsfile
RobertKrawitz: Yes, but if somebody manages to open the file in the (short) window between when you create it and when you chmod it, they could grab the value.  Make sure that you use `umask 77` while you create directories and files rather than relying on chmod after the fact.

RobertKrawitz (changes requested):

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
ebattat: This code is working properly more than 1 year, just move it to bare metal class.
I dont see any impact to return the missing name, It will require to hold the expected master/ worker nodes.

RobertKrawitz: It always helps to make it easier to debug.

benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
ebattat: Fixed, add dedicate cluster_type environment variable to support SNO

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
ebattat: Fixed

benchmark_runner/common/clouds/BM/bare_metal_exceptions.py
ebattat: Fixed

jenkins/PerfCI_OpenShift_Deployment/Jenkinsfile
ebattat: Fixed

jenkins/PerfCI_OpenShift_Deployment/Jenkinsfile
ebattat: I dont see any impact to use umask 77, I think its more clear to run chmod 600 after

jenkins/PerfCI_Operators_Deployment/Jenkinsfile
ebattat: If we put it in the secret we would know the current installed version, Are u sure its good to make it as secret ?

benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
ebattat: Fixed

benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
ebattat: Fixed

benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
ebattat: Fixed

benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
ebattat: Added

benchmark_runner/main/main.py
```diff
@@ -96,36 +100,58 @@ def install_ocp(ibm_operations: IBMOperations, step: str):
...
     install_resources_list = environment_variables_dict.get('install_resources_list', '')
     # convert str to list
     try:
         resources = ast.literal_eval(install_resources_list) if isinstance(install_resources_list, str) else install_resources_list
     except (ValueError, SyntaxError):
         resources = [install_resources_list]
-    logger.info(f'Start IBM OCP resources installation')
-    oc = ibm_operations.oc_login()
-    ibm_operations.verify_cluster_is_up(oc)
+    logger.info(f'Start BM OCP resources installation')
```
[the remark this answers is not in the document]
ebattat: Fixed

benchmark_runner/main/main.py
ebattat: Fixed

benchmark_runner/common/ocp_resources/odf/template/04_local_volume_set_template.yaml
RobertKrawitz: That's not ideal either, as it assumes specifics about our setup (what happens when somebody runs it on a very different cluster?).  But it's better.  Can't you inspect /dev/disk/... to determine what type of disk it is (NVMe vs. SATA)?

jenkins/PerfCI_OpenShift_Deployment/Jenkinsfile
RobertKrawitz: If you use `umask 77`, any files created will already be mode 600, so you don't need the explicit chmod.

jenkins/PerfCI_Operators_Deployment/Jenkinsfile
RobertKrawitz: Secret, environment variable, whatever.  Just not in code.

benchmark_runner/common/ocp_resources/odf/template/04_local_volume_set_template.yaml
ebattat: The only option is - NonRotational but its same for NVMe and SSD.
So the only option is by size or by vendor.
Here is the full yaml template for LocalVolumeSet:
```
apiVersion: local.storage.openshift.io/v1alpha1
kind: LocalVolumeSet
metadata:
  name: example-autodetect
spec:
  nodeSelector:
    nodeSelectorTerms:
      - matchExpressions:
          - key: kubernetes.io/hostname
            operator: In
            values:
              - worker-0
              - worker-1
  storageClassName: example-storageclass 
  volumeMode: Filesystem
  fsType: ext4
  maxDeviceCount: 10
  deviceInclusionSpec:
    deviceTypes: 
      - disk
      - part
    deviceMechanicalProperties:
      - NonRotational
    minSize: 10G
    maxSize: 100G
    models:
      - SAMSUNG
      - Crucial_CT525MX3
    vendors:
      - ATA
      - ST2000LM
```
For more details: https://docs.openshift.com/container-platform/4.14/storage/persistent_storage/persistent_storage_local/persistent-storage-local.html

@jeniferh, Do you think its safe enough to choose model and vendor for disks auto detect ?

jenkins/PerfCI_OpenShift_Deployment/Jenkinsfile
ebattat: ok so I will fix it to:
`umask 77 && mkdir -p $WORKSPACE/.ssh/`

jenkins/PerfCI_Operators_Deployment/Jenkinsfile
ebattat: Change it, it means that no need to open PR for upgrading new version anymore.
We need only update the secret, is that ok ?

jenkins/PerfCI_OpenShift_Deployment/Jenkinsfile
RobertKrawitz: Since the directory is created under umask 77 it will never be accessible, so it's OK for the private key underneath it to be not readable.  Best practice would be to use umask 77 for that too, but this is acceptable.

RobertKrawitz (approved):

benchmark_runner/common/ocp_resources/odf/template/04_local_volume_set_template.yaml
jeniferh: I think the filters are going to be very environmentally specific no matter what, including maxDeviceCount, although I suppose you can inspect the disks dynamically to grab the filter values. I've never actually used models or vendors, from the code it just says it needs to match what 'lsblk' provides. The other option instead of Discovery mode w/ LocalVolumeSet Filters is to just use LocalVolume and list each devicePath

PR #785, opened 2024-02-22
Add LSO node selector

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

There is currently an issue when there are more than 1 disk per node for LSO. 
The LSO node selector should address this by allocating a dedicated LSO node.

### DIFF
--- a/tests/integration/benchmark_runner/common/oc/test_oc.py
+++ b/tests/integration/benchmark_runner/common/oc/test_oc.py
@@ -107,7 +107,7 @@ def test_oc_get_free_disk_id():
     oc = OC(kubeadmin_password=test_environment_variable['kubeadmin_password'])
     oc.login()
     with mock.patch.object(OC, 'get_worker_disk_ids', new=mock_get_worker_disk_ids):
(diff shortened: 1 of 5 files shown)


tests/integration/benchmark_runner/common/oc/test_oc.py
```diff
@@ -107,7 +107,7 @@ def test_oc_get_free_disk_id():
     oc = OC(kubeadmin_password=test_environment_variable['kubeadmin_password'])
     oc.login()
     with mock.patch.object(OC, 'get_worker_disk_ids', new=mock_get_worker_disk_ids):
-        assert oc.get_free_disk_id()
+        assert oc.get_free_disk_id(node='worker-2')
```
RobertKrawitz, line 110: Rather than having `worker-2` scattered around the tests, this belongs in the tests environment variables so that you can change this in just one place.

ebattat: Fixed

PR #799, opened 2024-03-27
Update Grafana regression dashboard

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Update Grafana regression dashboard:
1. Remove Kata workloads
2. Update CI url
3. Update artifacts urls
4. Update metric: VM Memory => virt launcher


RobertKrawitz (commented):
Why are the new links http rather than https?  I realize that the new links are behind the firewall, but in general it's best to use https rather than http.

ebattat:
All those links are required VPN, not see any reason to use https

RobertKrawitz (approved):
Approved with the understanding that this will be made https when practicable.

PR #823, opened 2024-06-03
Python 3.11 and 3.12 support

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Adding Python 3.11/3.12 support
Update PyYAML==6.0 to yYAML==6.0.1 for support Python3.12

PR #831, opened 2024-06-04
Update python3.12 changes

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Update python 3.12 changes:
1. pip install setuptools before running: python3 setup.py --version
2. Update Dockerfile with python3.12
3. Update Dockerfile with latest virtctl/oc client versions
 
This PR must merged after [PR](https://github.com/redhat-performance/benchmark-runner/pull/823 ) because it used python3.12 packages dependencies inside Dockerfile

### DIFF
--- a/.github/workflows/Build_runner.yml
+++ b/.github/workflows/Build_runner.yml
@@ -26,10 +26,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - name: Set up Python 3.9
+    - name: Set up Python 3.12
       uses: actions/setup-python@v5
       with:
-        python-version: 3.9
+        python-version: 3.12
     - name: 🔨 Build and publish distribution 📦 to PyPI
       env:
         TWINE_USERNAME: __token__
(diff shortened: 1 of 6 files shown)


RobertKrawitz (changes requested):
832, 831, and 834 should be a single PR unless there's a specific reason why part needs to be in the repository before something else can go in.

.github/workflows/Build_runner.yml
```diff
@@ -59,8 +59,8 @@ jobs:
         # Verify and wait till latest benchmark-runner version will be updated in Pypi (timeout 900 seconds)
         timeout=900
         sleep_time=30
+        pip --no-cache-dir install setuptools benchmark-runner --upgrade
```
RobertKrawitz, line 62: Will this change what version shows up because you've moved this before expected_version rather than after, and is that an issue?

RobertKrawitz (commented):
Just one question.

.github/workflows/Build_runner.yml
ebattat: No, it fecthes the version from checkout repository setup file and from benckmark runner installed pkg version.

RobertKrawitz: OK

RobertKrawitz (approved):

ebattat:
Here are all the PRs that are involved in the Python 3.12 upgrade:
https://github.com/redhat-performance/benchmark-runner/pull/823
https://github.com/redhat-performance/benchmark-runner/pull/831
https://github.com/redhat-performance/benchmark-runner/pull/832
https://github.com/redhat-performance/benchmark-runner/pull/845

PR #843, opened 2024-06-06
How to use pre-commit

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [ ] enhancement
- [x] documentation
- [ ] dependencies

## Description

How to use pre-commit before every commit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,8 +13,9 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.4.0
     hooks:
+#- id: check-yaml # Comment it out because it affects all golden test YAML files.
       - id: check-json
-      - id: check-yaml
       - id: trailing-whitespace
+        exclude: benchmark_runner/grafana/perf/jsonnet/main.libsonnet
       - id: check-added-large-files
       - id: end-of-file-fixer
--- a/PRE-COMMIT.md
+++ b/PRE-COMMIT.md
@@ -1,0 +1,12 @@
+# How to use pre-commit
+
+## Install pre-commit
+```
+pip install pre-commit / dnf install pre-commit
+pre-commit install
+```
+
+## Run on every commit
+```
+pre-commit run --all-files
+```


PRE-COMMIT.md
```diff
@@ -0,0 +1,12 @@
+# How to use pre-commit
```
RobertKrawitz, line 1: This wasn't correct on my laptop.  `pip-install pre-commit` did not install a `pre-commit` command; I had to `dnf install pre-commit` to get the pre-commit command.

.pre-commit-config.yaml
```diff
@@ -13,8 +13,9 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.4.0
     hooks:
+#- id: check-yaml # Comment it out because it affects all golden test YAML files.
       - id: check-json
-      - id: check-yaml
       - id: trailing-whitespace
+        exclude: benchmark_runner/grafana/perf/jsonnet/main.libsonnet
```
RobertKrawitz, line 19: As main.libsonnet is a purely generated file, this makes sense.

.pre-commit-config.yaml
```diff
@@ -13,8 +13,9 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.4.0
     hooks:
+#- id: check-yaml # Comment it out because it affects all golden test YAML files.
```
RobertKrawitz, line 16: I'm not entirely sure about this.  The golden files are generated, but problems with them likely indicate problems with the templates used to generate them.  Do you have an example of where this causes problems?

RobertKrawitz (changes requested):
- Changes needed to PRE-COMMIT.md.
- I'm not convinced that disabling this check is the right way to approach this issue; the problem may be that the template files need to be fixed.

.pre-commit-config.yaml
ebattat: See all the errors in the attached file:
[precommit_yamls_errors.txt](https://github.com/user-attachments/files/15620071/precommit_yamls_errors.txt)

RobertKrawitz: I'm going to take a look at fixing those.

RobertKrawitz (commented):
I found a number of other useful hooks to add.  I'm going to open a new PR for all of that.

RobertKrawitz:
I've opened #844 to fix the check-yaml problems and added a number of new hooks in addition.

ebattat:
Merged in [PR](https://github.com/redhat-performance/benchmark-runner/pull/844)

ebattat:
Merged in [PR](https://github.com/redhat-performance/benchmark-runner/pull/844)

PR #849, opened 2024-06-19
Add delete s3 key test

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Add delete s3 key test and did some small fixes

### DIFF
--- a/benchmark_runner/common/clouds/shared/s3/s3_operations.py
+++ b/benchmark_runner/common/clouds/shared/s3/s3_operations.py
@@ -38,7 +38,7 @@ class S3Operations:
     @typeguard.typechecked
     def upload_file(self, file_name_path: str, bucket: str, key: str, upload_file: str):
         """
-        This method upload file to s3
+        This method uploads file/object to s3
         :param file_name_path:'/home/user/test.txt'
         :param bucket:'benchmark'
         :param key:'test-data'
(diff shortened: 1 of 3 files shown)


benchmark_runner/common/clouds/shared/s3/s3_operations.py
```diff
@@ -153,10 +153,29 @@ def file_exist(self, bucket: str, key: str, file_name: str):
...
+        :return:
+        """
+        try:
+            response = self.__s3_client.list_objects_v2(Bucket=bucket, Prefix=key)
+            if response.get('Contents'):
+                return True
+            return False
+        # Todo add custom error
+        except ClientError:
+            raise
```
RobertKrawitz, line 171: This is valid but confusing.  Is this specifically for the client error?  This would be more obvious (IMO):

```
except ClientError as exc:
    raise exc from null
```

But if you want to keep it this way I won't seriously object.

RobertKrawitz (approved):

benchmark_runner/common/clouds/shared/s3/s3_operations.py
ebattat: This is the way to throw an exception to the upper level.

PR #855, opened 2024-06-26
Update PerfCi to 4.16.0 release

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

There is a new release for [4.16](https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/releasestream/4-stable/release/4.16.0)
Update PerfCi from the latest 4.16.0-rc to the 4.16.0 release

### DIFF
--- a/.github/workflows/Weekly_Perf_Env_Installer_CI.yml
+++ b/.github/workflows/Weekly_Perf_Env_Installer_CI.yml
@@ -56,8 +56,8 @@ jobs:
           END
       - name: ▶ OCP assisted installer
         env:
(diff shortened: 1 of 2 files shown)


jeniferh (approved):
LGTM 🥳

RobertKrawitz (approved):
Curious why the OCP_CLIENT_VERSION differed between the dockerfile and the workflow.  Doesn't really matter since they're now the same.  Approving this.

It would be good in general if all of these variables came from one place rather than various identical YAML fragments everywhere.  But that's a subject for another day.

PR #860, opened 2024-08-02
Multiprocessing support for MAC

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Currently, there is no option to run multiprocessing tasks on a Mac without issues. 
Adding set_start_method('fork') is necessary because the default method is spawn.
[For more details](https://docs.python.org/3/library/multiprocessing.html#:~:text=The%20default%20on%20Windows%20and%20macOS.)

### DIFF
--- a/benchmark_runner/main/main.py
+++ b/benchmark_runner/main/main.py
@@ -1,6 +1,8 @@
 
 import ast  # change string list to list
 from enum import Enum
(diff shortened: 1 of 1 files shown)


athiruma (approved):
/lgtm

PR #861, opened 2024-08-05
Add VM Scale Management Steps

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

[Chaos Testing] Adding VM Scale Management Steps:
- Run VMs without deleting them before cluster upgrade (Add DELETE_ALL variable)
- Verify VMs only after cluster upgrade (Add VERIFICATION_ONLY variable)
- Add vm_ssh field to verify that SSH into VM is working properly

### DIFF
--- a/benchmark_runner/workloads/bootstorm_vm.py
+++ b/benchmark_runner/workloads/bootstorm_vm.py
@@ -5,6 +5,7 @@ from multiprocessing import Process
 
 from benchmark_runner.common.logger.logger_time_stamp import logger_time_stamp, logger
 from benchmark_runner.common.elasticsearch.elasticsearch_exceptions import ElasticSearchDataNotUploaded
+from benchmark_runner.workloads.workloads_exceptions import MissingVMs
 from benchmark_runner.workloads.workloads_operations import WorkloadsOperations
 from benchmark_runner.common.oc.oc import VMStatus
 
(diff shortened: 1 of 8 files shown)


benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -123,14 +134,39 @@ def _run_vm(self):
...
             yaml=os.path.join(f'{self._run_artifacts_path}', f'{self._name}.yaml'),
             vm_name=self._vm_name)
 
+    def _verify_vm_ssh(self):
+        """
+        This method verifies each VM ssh login
+        :return:
+        """
+        try:
+            vm_names = self._oc._get_all_vm_names()
```
RobertKrawitz, line 152: It might make sense to compare this against a list of expected VM names, if that's practical here.  And you certainly want to make sure that there is at least one, or you'll simply be "verifying" a null case.

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -66,23 +66,33 @@ def _set_bootstorm_vm_start_time(self, vm_name: str = ''):
...
+            # wait till vm ssh login
+            if self._oc.get_vm_node(vm_name=vm_name):
+                vm_node = self._oc.get_vm_node(vm_name=vm_name)
+                if vm_node:
+                    node_ip = self._oc.get_nodes_addresses()[vm_node]
+                    vm_node_port = self._oc.get_exposed_vm_port(vm_name=vm_name)
+                    if self._oc.wait_for_vm_ssh(vm_name=vm_name, node_ip=node_ip, vm_node_port=vm_node_port):
+                        logger.info(f"Successfully ssh into VM: '{vm_name}' in Node: '{vm_node}' ")
+                    return vm_node
+            return False
```
RobertKrawitz, line 84: Make sure you capture any stderr from the ssh command, so that when it fails you can at least have some hope of figuring out what went wrong (it might simply be a network problem or a credential problem).

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -103,8 +113,9 @@ def _finalize_vm(self):
         metric_results = self._prometheus_metrics_operation.run_prometheus_queries()
         prometheus_result = self._prometheus_metrics_operation.parse_prometheus_metrics(data=metric_results)
         # update total vm run time
-        total_run_time = self._get_bootstorm_vm_total_run_time()
-        self._data_dict.update({'total_run_time': total_run_time})
+        if not self._verification_only:
+            total_run_time = self._get_bootstorm_vm_total_run_time()
+            self._data_dict.update({'total_run_time': total_run_time})
```
RobertKrawitz, line 118: It wouldn't be a bad idea to stick some distinctive value in there even in verification mode, such as -1, so that the schema is the same both ways.

RobertKrawitz (changes requested):

benchmark_runner/workloads/bootstorm_vm.py
ebattat: total_run_time will be empty when no input data

benchmark_runner/workloads/bootstorm_vm.py
ebattat: You right, add MissingVMs error when no running VMs at all

benchmark_runner/workloads/bootstorm_vm.py
RobertKrawitz: Does that actually capture the stderr from the ssh command?

benchmark_runner/workloads/bootstorm_vm.py
RobertKrawitz: Understood, but that's about testing that total_run_time is handled correctly.  Not that big of a deal, but think about it.

benchmark_runner/workloads/bootstorm_vm.py
ebattat: You can see the full code [here](https://github.com/redhat-performance/benchmark-runner/blob/4e9c29f8977649c9c13659779e87a2095d800c36/benchmark_runner/common/ssh/ssh.py#L24-L41)

benchmark_runner/workloads/bootstorm_vm.py
ebattat: I prefer that it will be empty in ElasticSearch instead of -1, more simple to handle it.

ebattat:

benchmark_runner/workloads/bootstorm_vm.py
ebattat: We have a custom error for it:
raise VMStateTimeout(vm_name=vm_name, state='ssh')

RobertKrawitz: But does it report the actual error reported by ssh?  I want that to be reported to make it easier to debug and to distinguish glitches from real problems.  It doesn't look like VMStateTimeout captures the reason for the error (what ssh reports on stderr).

ebattat:
@RobertKrawitz, we dont need to get ssh output, we need to know if vm is accessible for ssh call or not

RobertKrawitz:
> @RobertKrawitz, we dont need to get ssh output, we need to know if vm is accessible for ssh call or not

If the ssh fails, we want to know why.  Connection timed out, connection reset by peer, authentication are different failures that ssh can report and I believe we do want to see what happened.

ebattat:
If something will fail, I will raise it [here](https://github.com/redhat-performance/benchmark-runner/blob/3f556a5290006a3bcb94447bfd64af2454090d38/benchmark_runner/common/ssh/ssh.py#L40)

RobertKrawitz:
> If something will fail, I will raise it [here](https://github.com/redhat-performance/benchmark-runner/blob/3f556a5290006a3bcb94447bfd64af2454090d38/benchmark_runner/common/ssh/ssh.py#L40)

As long as stderr from ssh gets saved, I don't care how you do it.  I do want to be sure it's possible to see after the fact what happened.

ebattat:
Sorry I didnt get your point.

RobertKrawitz:
> Sorry I didnt get your point.

I want any stderr output from ssh to be saved, so that if there is a failure, we can look at it to try to figure out what happened.  ssh can fail for any number of reasons which may have nothing to do with whether the VM is running: there might be an authentication failure, there might be a transient network problem, there might be any number of things.  Simply "ssh to node <whatever> failed" doesn't help us figure out what happened.

ebattat:
I am focusing on the main flow right now and I dont want to focus in log collection in this pr only on verification steps.
I will open a separate PR for VM logs collection.

RobertKrawitz:
> I am focusing on the main flow right now and I dont want to focus in log collection in this pr only on verification steps. I will open a separate PR for VM logs collection.

OK, but I think that it's important to collect information if verification fails.  I will accept opening a PR for that, but I'd like you to open an issue against it prior to my approving this PR.

ebattat:
Open issue regarding it: https://github.com/redhat-performance/benchmark-runner/issues/862

PR #864, opened 2024-08-20
Stable Operators Deployment for OCP Upgrade 

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. Add the ability to install stable operators: LSO, ODF, and CNV for OCP upgrades.
2. Update LSO/ODF to install from the stable channel and use installPlanApproval: Automatic.
3. Add a wait_for_csv method with number validation and add a custom error: CSVNotCreateTimeout.
4. Implement a double-check for operator installation validation, which is mandatory for upgrades between operator internal releases.
5. If the operator installation fails, trigger the PerfCI_OpenShift_Deployment Jenkinsfile.


benchmark_runner/common/oc/oc.py
```diff
@@ -654,6 +654,27 @@ def wait_for_initialized(self, label: str, workload: str = '', status: str = 'In
...
+        :param csv_num: number of csvs, default 1
+        :param namespace:
+        :param timeout:
+        :return: csv names if getting csv or raise CSVError
+        """
+        current_wait_time = 0
+        while timeout <= 0 or current_wait_time <= timeout:
+            csv_names = self.run(f"{self.__cli} get csv -n {namespace} -o jsonpath={{$.items[*].metadata.name}}")
+            if csv_names and len(csv_names.split()) == csv_num:
+                return csv_names
```
RobertKrawitz, line 672: Need to handle the case of getting more CSVs than you expect.

Also, do you care what phase they're in at this point?

benchmark_runner/common/ocp_resources/create_odf.py
```diff
@@ -52,14 +54,9 @@ def create_odf(self):
                                                       verify_cmd=r"""oc get pv -o jsonpath="{range .items[*]}{.metadata.name}{'\n'}{end}" | grep local | wc -l""",
                                                       count_openshift_storage=True)
                 if '07_subscription.yaml' in resource:
-                    # wait till get the patch
-                    self.wait_for_ocp_resource_create(resource='odf',
-                                                      verify_cmd="oc get InstallPlan -n openshift-storage -ojsonpath={.items[0].metadata.name}",
-                                                      status="install-")
-                    self.apply_patch(namespace='openshift-storage', resource='odf')
-                elif '08_storage_cluster.yaml' in resource:
                     # Must be run after installing the storage cluster because CSVs sometimes fail
```
RobertKrawitz, line 57: Looks like this comment has effectively moved from 08_storage_cluster.yaml to 07_subscription.yaml.

RobertKrawitz (changes requested):

benchmark_runner/common/oc/oc.py
ebattat: csv_num is a dynamically number according to relevant operator, the default is 1

benchmark_runner/common/ocp_resources/create_odf.py
ebattat: Yes, in the past we used installPlanApproval: Manual and I have changed it to installPlanApproval: Automatic, that cause that issue, remove this comment.

benchmark_runner/common/oc/oc.py
RobertKrawitz: What happens if `oc get csv...` returns _more_ CSVs than what you expect, but includes everything you do want?

ebattat: I can change it to:
`if csv_names and len(csv_names.split()) >= csv_num:`
WDYT ?

RobertKrawitz: I think ideally you'd check to make sure that you have all of the specific CSVs that you want, but short of that, I think >= would be a better test.

ebattat: Done

RobertKrawitz (approved):

PR #901, opened 2024-09-26
Enable Google Drive  & Disable Prometheus during VM validation

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. Disable Prometheus during VM verifications - no need to collects prometheus logs during VM verifications.
We already collect each VM Prometheus when creating it.
3. Enable Google drive for storing logs in case of errors


ebattat:
@RobertKrawitz, updated the description

RobertKrawitz (approved):

PR #902, opened 2024-09-26
Wait for install complete

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Currently, the installation shell script finishes without validating the completion of the installation. Add a wait step to ensure the installation is complete.
[see logs](https://github.com/redhat-performance/benchmark-runner/actions/runs/11055021914/job/30713416342)


RobertKrawitz (approved):
It appears that Azure does not have a `run_ocp_installer` method, so this is in fact complete at this time.

PR #905, opened 2024-09-27
Paramiko > 3.4.1 cause to SSH session not active

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Paramiko 3.4.0 causes the following warning:
`/root/benchmark-runner/venv/lib/python3.12/site-packages/paramiko/pkey.py:100: CryptographyDeprecationWarning: TripleDES has been moved to cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES and will be removed from this module in 48.0.0.`

After upgrading to Paramiko 3.4.1/3.5.0, the following error occurs when running a remote shell:
 Got [SSH session not active](https://github.com/redhat-performance/benchmark-runner/actions/runs/11068161437/job/30753164624)

When revert back to paramiko==3.4.0 [all working properly](https://github.com/redhat-performance/benchmark-runner/actions/runs/11072629312/job/30767282287).

Currently, I am using filterwarnings to temporarily suppress the warning and adding a TODO to follow up on the Paramiko fix.

### DIFF
--- a/benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
+++ b/benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
@@ -228,7 +228,6 @@ class BareMetalOperations:
         logger.info(f'Starting OCP assisted installer, Start time: {datetime.now().strftime(datetime_format)}')
         # Must add -t otherwise remote ssh of ansible will not end
         self._ssh.run(f"ssh -t provision \"{self._install_ocp_cmd()}\" ")
-        self._wait_for_install_complete()
         logger.info(f'OpenShift cluster {self._get_installation_version()} version is installed successfully, End time: {datetime.now().strftime(datetime_format)}')
 
     @logger_time_stamp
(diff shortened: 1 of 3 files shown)


RobertKrawitz:
I suspect that you're using an old ssh key based on an outdated algorithm.  You should update the key.

ebattat:
@RobertKrawitz,
Our Test env was down for 2 weeks due to this issue.
After revert paramiko from 3.4.1 -> 3.4.0 issue was solved
[Installation passed successfully](https://github.com/redhat-performance/benchmark-runner/actions/runs/11074252176)
Do you think we should open an issue in [paramiko project](https://github.com/paramiko/paramiko/issues) regarding it ?

RobertKrawitz:
You can try, but I suspect that the issue is that the ssh key you're using relies on an outdated algorithm and that it's behaving as intended.  Are you using a fixed ssh key from somewhere?

ebattat:
@RobertKrawitz,
Issue was solved when revert back to paramiko 3.4.0.
Open paramiko [issue](https://github.com/paramiko/paramiko/issues/2459)

darouwan:
Meet the same issue. So It's solved in 3.5.0?

ebattat:
Hi @darouwan,

No, we are still using [3.4.0](https://github.com/redhat-performance/benchmark-runner/blob/de15a80467528856e60731bf14f5cfe2a7996f87/requirements.txt#L17)

PR #908, opened 2024-09-30
Verify Node Ready

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Check the node status after the OpenShift installation or upgrade, and raise an error if any node is not in 'Ready' status. Additionally, add relevant unit tests or integration tests.

### DIFF
--- a/benchmark_runner/common/oc/oc.py
+++ b/benchmark_runner/common/oc/oc.py
@@ -11,7 +11,7 @@ from benchmark_runner.common.oc.oc_exceptions import (PodNotCreateTimeout, PodNo
     PodNotCompletedTimeout, PodTerminateTimeout, PodNameNotExist, LoginFailed, VMNotCreateTimeout, VMDeleteTimeout, \
     YAMLNotExist, VMNameNotExist, VMNotInitializedTimeout, VMNotReadyTimeout, VMStateTimeout, VMNotCompletedTimeout, \
     ExecFailed, PodFailed, DVStatusTimeout, CSVNotCreateTimeout, UpgradeNotStartTimeout, OperatorInstallationTimeout, \
-    OperatorUpgradeTimeout, ODFHealthCheckTimeout)
+    OperatorUpgradeTimeout, ODFHealthCheckTimeout, NodeNotReady)
 from benchmark_runner.common.ssh.ssh import SSH
 from benchmark_runner.main.environment_variables import environment_variables
 
@@ -375,7 +375,6 @@ class OC(SSH):
         else:
             raise PodNotReadyTimeout(label)
 
-
     def wait_for_odf_healthcheck(self, pod_name: str, namespace: str,
                                  timeout: int = SHORT_TIMEOUT):
         """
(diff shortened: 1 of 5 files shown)


benchmark_runner/common/oc/oc.py
```diff
@@ -441,6 +440,33 @@ def get_worker_nodes(self):
...
+
+    def verify_nodes_ready(self):
+        """
+        This method verifies that all nodes are in 'Ready' status.
+        If any node is not ready, it raises an error with the node name and its status.
+        @return: None
+        """
+        # Get the node name and status for all nodes
+        nodes_list = self.run(f"{self.__cli} get nodes --no-headers | awk '{{print $1, $2}}'").splitlines()
+        return self.check_node_status(nodes_list)
```
RobertKrawitz, line 468: The comment says that there is no return, but it does in fact return a value.

benchmark_runner/common/oc/oc.py
```diff
@@ -441,6 +440,33 @@ def get_worker_nodes(self):
...
+        @return:
+        """
+        # Check if any node is not in 'Ready' status
+        for node in nodes_list:
+            node_name, node_status = node.split()
+            if node_status != 'Ready':
+                raise NodeNotReady(node_name, node_status)
+
+        # If no nodes are found in a non-ready state
+        return "All nodes are ready."
```
RobertKrawitz, line 458: Do you need to do this, or can you just return True?

Also, rather than raising an exception from here, perhaps return False, or a list of nodes that are not ready so the caller could decide what to do?

RobertKrawitz (changes requested):

benchmark_runner/common/oc/oc.py

ebattat:

benchmark_runner/common/oc/oc.py
```diff
@@ -441,6 +440,33 @@ def get_worker_nodes(self):
...
+        """
+        This method check node status
+        @param nodes_list:
+        @return: True when all nodes in ready status
+        """
+        # Check if any node is not in 'Ready' status
+        for node in nodes_list:
+            node_name, node_status = node.split()
+            if node_status != 'Ready':
+                raise NodeNotReady(node_name, node_status)
```
RobertKrawitz, line 455: Cleaner to return False here, and let the caller decide whether to raise an exception.

benchmark_runner/common/oc/oc_exceptions.py
```diff
@@ -164,3 +164,10 @@ class ODFHealthCheckTimeout(OCError):
     def __init__(self):
         self.message = f"ODF health check timeout"
         super(ODFHealthCheckTimeout, self).__init__(self.message)
+
+
+class NodeNotReady(OCError):
+    """This exception return odf healthcheck timeout error"""
```
RobertKrawitz, line 170: "returns"

(yeah, have to like how an "s" suffix in English is normally the plural for nouns but the singular for verbs and vice versa)

benchmark_runner/common/oc/oc.py
ebattat: I think we must raise error if one of the node is not ready and CI failure

PR #909, opened 2024-09-30
Skip CNV/ODF upgrade when Cluster is already upgraded

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Skip CNV/ODF upgrade when Cluster is already upgraded

### DIFF
--- a/benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
+++ b/benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
@@ -230,24 +230,32 @@ class BareMetalOperations:
         self._ssh.run(f"ssh -t provision \"{self._install_ocp_cmd()}\" ")
         logger.info(f'OpenShift cluster {self._get_installation_version()} version is installed successfully, End time: {datetime.now().strftime(datetime_format)}')
 
-    @logger_time_stamp
-    def run_ocp_upgrade(self, oc: OC):
+    def is_ocp_already_upgraded(self, oc: OC):
         """
-        This method runs ocp upgrade
-        @param oc:
-        :return: True if installation success and raise exception if installation failed
+        This method checks if Openshift cluster is already upgraded
+        @return:
         """
         if f'Cluster version is {self._upgrade_ocp_version}' == oc.get_cluster_status():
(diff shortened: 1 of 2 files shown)


benchmark_runner/common/clouds/BareMetal/bare_metal_operations.py
```diff
@@ -239,6 +239,7 @@ def run_ocp_upgrade(self, oc: OC):
         """
         if f'Cluster version is {self._upgrade_ocp_version}' == oc.get_cluster_status():
             logger.info(f'Cluster is already upgraded to: {self._upgrade_ocp_version}')
+            return 'Cluster is already upgraded'
```
RobertKrawitz, line 242: Should return True, not a string (or you need to change the comment above)..  Also note that there's no return if the version is not correct (existing code, but it looks wrong).

benchmark_runner/main/main.py
```diff
@@ -159,10 +159,10 @@ def upgrade_ocp_bare_metal(step: str):
     bare_metal_operations.connect_to_provisioner()
     oc = bare_metal_operations.oc_login()
     if step == 'run_bare_metal_ocp_upgrade':
-        bare_metal_operations.run_ocp_upgrade(oc)
-        # The LSO/ODF upgrade must be run manually after the OCP upgrade for the channel version; it won't upgrade automatically.
-        bare_metal_operations.install_ocp_resources(resources=['lso'], upgrade_version=lso_version)
-        bare_metal_operations.install_ocp_resources(resources=['odf'], upgrade_version=odf_version)
+        if not bare_metal_operations.run_ocp_upgrade(oc) == 'Cluster is already upgraded':
```
RobertKrawitz, line 162: Is there anything you need to do if the status is anything different (e. g. if either the upgrade fails or the upgrade does something, so it's not previously upgraded)?

RobertKrawitz (changes requested):

benchmark_runner/main/main.py
ebattat: Add dedicate method is_ocp_already_upgraded

ebattat:

RobertKrawitz (approved):

PR #910, opened 2024-09-30
WAIT FOR UPGRADE VERSION

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Currently, the stage 'WORKLOADS Verifications' runs until the stage 'OpenShift Upgrade AND Verifications' is complete. Change it to use WAIT_FOR_UPGRADE_VERSION inside the stage 'WORKLOADS Verifications', so that it runs the Windows verification in a single job loop.

### DIFF
--- a/jenkins/PerfCI_Chaos/04_PerfCI_Chaos_Upgrade_OpenShift_Deployment/Jenkinsfile
+++ b/jenkins/PerfCI_Chaos/04_PerfCI_Chaos_Upgrade_OpenShift_Deployment/Jenkinsfile
@@ -151,8 +151,6 @@ END
                                                             --privileged "${QUAY_BENCHMARK_RUNNER_REPOSITORY}"
 
                                                     """
-                                                   // Set the flag to indicate that upgrade steps are completed
-                                                   currentBuild.description = "UpgradeStageCompleted"
                                                 }
                                             }
                                         }
@@ -172,12 +170,8 @@ END
                     def WINDOWS_URL = ''
                     def SCALE = ''
                     def workload_name = ''
-                    def retryCount = 3 // Number of retries per workload
-                    // while
-                    while (currentBuild.description != "UpgradeStageCompleted") {
                     for (workload in workloads) {
-                        boolean success = false
-                        for (int i = 0; i < retryCount && !success; i++) {
+                            boolean success = false
                             try {
                                 // workload full name
                                 workload_name = workload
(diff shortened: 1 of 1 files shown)


jenkins/PerfCI_Chaos/04_PerfCI_Chaos_Upgrade_OpenShift_Deployment/Jenkinsfile
```diff
@@ -264,14 +258,12 @@ END
                                 } else {
                                     echo "Reached maximum retry attempts for workload ${workload}. Moving to the next workload..."
                                 }
-                            }
```
RobertKrawitz, line 267: Above here, do you need to check that the `sudo podman run` succeeds before `success = true`?

RobertKrawitz (commented):
My comment isn't about the changes per se, but I want to determine whether the `podman run` should have its status checked before setting success to true.

jenkins/PerfCI_Chaos/04_PerfCI_Chaos_Upgrade_OpenShift_Deployment/Jenkinsfile
ebattat: I dont think it necessary

RobertKrawitz: Why?  It's a simple check, and things can go wrong.

ebattat: Its only on failure

RobertKrawitz: It looks like you're relying on something raising an excception if something goes wrong, but what if the `podman run` simply fails?  Shouldn't that be treated as a failure?

ebattat: If podman run will fail, I prefer that all the run the will fail.

RobertKrawitz: Right, but that's why you should check whether the podman run fails -- if it fails, then you fail the entire run.

PR #911, opened 2024-10-01
Hot Fix: Sleep 1 second between each VMs verification cycle

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Sleep 1 second between each VMs verification cycle.
That will cover VMs verification during Chaos testing.

### DIFF
--- a/benchmark_runner/workloads/bootstorm_vm.py
+++ b/benchmark_runner/workloads/bootstorm_vm.py
@@ -196,7 +196,9 @@ class BootstormVM(WorkloadsOperations):
                             failure = True
                         upgrade_done = self._oc.get_cluster_status() == f'Cluster version is {self._wait_for_upgrade_version}'
 
-                    current_wait_time += self._oc.SHORT_TIMEOUT # Increment the wait time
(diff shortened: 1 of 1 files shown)


benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -196,7 +196,9 @@ def _verify_vm_ssh(self):
                             failure = True
                         upgrade_done = self._oc.get_cluster_status() == f'Cluster version is {self._wait_for_upgrade_version}'
 
-                    current_wait_time += self._oc.SHORT_TIMEOUT # Increment the wait time
+                    # Sleep 1 sec between each cycle
+                    time.sleep(1)
+                    current_wait_time += 1  # Increment the wait time
```
RobertKrawitz, line 201: Ideally you'd check the actual time rather than just incrementing the current_wait_time, but for this purpose where the exact timeout doesn't matter, this is OK.

RobertKrawitz (approved):

PR #916, opened 2024-10-10
Check the VM status using virtctl instead of the SSH protocol

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Change the VM login method from SSH protocol to Virtctl, as this will provide clearer error messages.
Additionally, Since the VM could be in the middle of a migration, add a retry mechanism that makes 5 attempts with a 10-second pause between each retry.

### DIFF
--- a/benchmark_runner/workloads/bootstorm_vm.py
+++ b/benchmark_runner/workloads/bootstorm_vm.py
@@ -68,21 +68,21 @@ class BootstormVM(WorkloadsOperations):
 
     @logger_time_stamp
     def _wait_ssh_vm(self, vm_name: str):
-            """
-            Verify ssh into VM and return vm node in success or False if failed
-            @return:
-            """
-            self._virtctl.expose_vm(vm_name=vm_name)
-            # wait till vm ssh login
-            if self._oc.get_vm_node(vm_name=vm_name):
-                vm_node = self._oc.get_vm_node(vm_name=vm_name)
-                if vm_node:
-                    node_ip = self._oc.get_nodes_addresses()[vm_node]
-                    vm_node_port = self._oc.get_exposed_vm_port(vm_name=vm_name)
-                    if self._oc.wait_for_vm_ssh(vm_name=vm_name, node_ip=node_ip, vm_node_port=vm_node_port):
-                        logger.info(f"Successfully ssh into VM: '{vm_name}' in Node: '{vm_node}' ")
-                    return vm_node
-            return False
+        """
+        This method verifies ssh into VM and return vm node in success or False if failed
+        @return:
+        """
+        self._virtctl.expose_vm(vm_name=vm_name)
+        # wait till vm ssh login
+        if self._oc.get_vm_node(vm_name=vm_name):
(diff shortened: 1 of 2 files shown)


benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -149,47 +149,49 @@ def _run_vm(self):
...
             try:
-                vm_node = self._oc.get_vm_node(vm_name)
-                node_ip = self._oc.get_nodes_addresses().get(vm_node)
-
-                if node_ip:
-                    vm_node_port = self._oc.get_exposed_vm_port(vm_name)
-                    ssh_status = self._oc.get_vm_ssh_status(vm_name, node_ip, vm_node_port)
-                    vm_ssh = 1 if ssh_status == 'True' else 0
+                ssh_status = self._oc.get_vm_virtctl_status(vm_name)
+                vm_ssh = 1 if ssh_status == 'True' else 0
```
RobertKrawitz, line 164: Why do you need this separately from `ssh_status`?  You can always use `int(vm_ssh)` for the same purpose.

ebattat: In case of error we return the error message:
```
virtualmachineinstance.kubevirt.io "windows-vm-65265214-0" not found
kex_exchange_identification: Connection closed by remote host
```

RobertKrawitz: That's fine, but why does `vm_ssh` need to be distinct from `ssh_status`?

RobertKrawitz: Oh, I see...it would be cleaner to return the boolean status and the message as two return values:

```
if ok:
    return (True, '')
else:
    return (False, 'This is why it failed')

ebattat: We need 2 parameters:
1.  vm_ssh: True/False - if ssh passed successfully (Filtering in Grafana only error msg)
2.  ssh_status: error msg in case of error

RobertKrawitz: Exactly; don't overload them.

ebattat: I prefer to keep it separately for Grafana dashboard

RobertKrawitz: Can't you convert it only when you need to insert it into grafana, rather than carrying around two separate variables that could get out of sync?

ebattat: @RobertKrawitz I agree.
Use the 'virtctl_status' string variable (instead of 'ssh_status') with the value 'True' if successful, or an error message if it fails.
Additionally, add the 'virtctl_vm' integer value (instead of 'vm_ssh') only for Grafana.

ebattat:
@RobertKrawitz,
Any more comments ?

RobertKrawitz (approved):

PR #920, opened 2024-10-16
Missing Setup Python

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Got the following error:
```
Run python -m pip install --upgrade pip
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP [6](https://github.com/redhat-performance/benchmark-runner/actions/runs/11362505417/job/31608260582#step:3:7)68 for the detailed specification.
```
[For more details](https://github.com/redhat-performance/benchmark-runner/actions/runs/11362505417/job/31608260582)

### DIFF
--- a/.github/workflows/Perf_Env_Build_Test_CI.yml
+++ b/.github/workflows/Perf_Env_Build_Test_CI.yml
@@ -406,6 +406,10 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip


.github/workflows/Perf_Env_Build_Test_CI.yml
```diff
@@ -408,7 +408,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Install dependencies
```
athiruma, line 409:
```
- name: Set up Python ${{ matrix.python-version }}
  uses: actions/setup-python@v5
  with:
      python-version: ${{ matrix.python-version }}
```

You missed this, It might be the reason for this error.

.github/workflows/Perf_Env_Build_Test_CI.yml
```diff
@@ -408,7 +408,7 @@ jobs:
...
         run: |
-          python -m pip install --upgrade pip
+          python -m pip install --upgrade pip --break-system-packages
```
athiruma, line 411: If we use python setup-python we don't require `--break-system-packages`.

.github/workflows/Perf_Env_Build_Test_CI.yml
ebattat: Right, fixed

athiruma (approved):
/lgtm

PR #924, opened 2024-11-03
Wait until the nodes are ready

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Currently, there is only node verification without waiting for it to reach the ready status. 
Adding a wait for the node to be ready until timeout.

### DIFF
--- a/benchmark_runner/common/oc/oc.py
+++ b/benchmark_runner/common/oc/oc.py
@@ -457,32 +457,61 @@ class OC(SSH):
         """
         return self.run(fr""" {self.__cli} get nodes -l node-role.kubernetes.io/worker= -o jsonpath="{{range .items[*]}}{{.metadata.name}}{{'\n'}}{{end}}" """)
 
-    @staticmethod
     @typechecked
-    def check_node_status(nodes_list: list):
+    def wait_for_node_ready(self, node: str = None, wait_time: int = None, timeout: int = int(environment_variables.environment_variables_dict['timeout'])):
         """
-        This method check node status
-        @param nodes_list:
-        @return: True when all nodes in ready status
+        This method waits until all nodes are in 'Ready' status or specific node
+        @param node: wait for specific node to be ready, when None check all nodes
+        @param wait_time: wait time between each loop
+        @param timeout: Maximum wait time in seconds, negative value means no timeout (default set in environment variables)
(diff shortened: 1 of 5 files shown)


benchmark_runner/common/oc/oc.py
```diff
@@ -457,32 +457,51 @@ def get_worker_nodes(self):
...
     @typechecked
-    def check_node_status(nodes_list: list):
+    def wait_for_nodes_ready(self, wait_time: int = None, timeout: int = int(environment_variables.environment_variables_dict['timeout'])):
         """
-        This method check node status
-        @param nodes_list:
-        @return: True when all nodes in ready status
+        This method waits until all nodes are in 'Ready' status.
+        @param wait_time: wait time between each loop
+        @param timeout: Maximum wait time in seconds (default set in environment variables)
```
RobertKrawitz, line 465: Need to document that negative value means no timeout (which I think is a good thing to have).

benchmark_runner/common/oc/oc.py
```diff
@@ -457,32 +457,51 @@ def get_worker_nodes(self):
...
+        This method checks the status of all nodes in the list.
+        @return: True if all nodes are in 'Ready' status; otherwise, return (node_name, node_status)
         """
         # Check if any node is not in 'Ready' status
-        for node in nodes_list:
+        for node in self.get_node_status():
             node_name, node_status = node.split()
             if node_status != 'Ready':
-                raise NodeNotReady(node_name, node_status)
+                return node_name, node_status
```
RobertKrawitz, line 492: Return all nodes that are not ready, not just the first one.

RobertKrawitz (changes requested):
One nit.

benchmark_runner/common/oc/oc.py
ebattat: Good catch, add dictionary to hold not ready node: `not_ready_nodes = {}`

benchmark_runner/common/oc/oc.py
ebattat: Add it to the document

RobertKrawitz (approved):

PR #934, opened 2024-11-28
Read Google Drive credentials from a file instead of a string

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Currently, Google Drive credentials are read from a string instead of a file, which causes an error when accessing them from the Jenkinsfile. 
This issue can be resolved by reading the credentials from a secret file instead of a secret string.


ebattat:
@RobertKrawitz, any comments ?

PR #943, opened 2024-12-12
Mount google drive credential files from volume

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Currently, we pass the credentials file path without a volume mount. We are adding it to ensure it is passed correctly.

### DIFF
--- a/jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile
+++ b/jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile
@@ -21,6 +21,9 @@ pipeline {
         GOOGLE_DRIVE_PATH = credentials('perfci_google_drive_path')
         GOOGLE_DRIVE_CREDENTIALS_FILE = credentials('perfci_google_drive_credentials_file')
         GOOGLE_DRIVE_TOKEN_FILE = credentials('perfci_google_drive_token_file')
+        CREDENTIALS_FILE = credentials('perfci_credentials_file')
+        TOKEN_FILE = credentials('perfci_token_file')
+        DESTINATION_PATH = credentials('perfci_destination_path')
         GOOGLE_DRIVE_SHARED_DRIVE_ID = credentials('perfci_google_drive_shared_drive_id')
         RUN_ARTIFACTS_URL = credentials('perfci_run_artifacts_url')
         REDIS = credentials('perfci_redis')
@@ -106,6 +109,33 @@ END
                     '''
                     sh 'sudo chmod 600 $CONFIG_PATH'
                 }
+            }
+        }
+
+        stage('Prepare Google credential files') {
+            steps {
+                    script {
+                        // Assign the file path to the environment variable
+                        env.GOOGLE_DRIVE_CREDENTIALS = "${env.WORKSPACE}/${CREDENTIALS_FILE}"
+                        env.GOOGLE_DRIVE_TOKEN = "${env.WORKSPACE}/${TOKEN_FILE}"
+
+                        sh """
+                        echo "Checking if ${env.GOOGLE_DRIVE_CREDENTIALS} exists..."
+                        if [[ -f "${env.GOOGLE_DRIVE_CREDENTIALS}" ]]; then
+                            echo "File exists. Deleting it..."
+                            rm -f "${env.GOOGLE_DRIVE_CREDENTIALS}"
+                        fi
+                        echo "Checking if ${env.GOOGLE_DRIVE_TOKEN} exists..."
+                        if [[ -f "${env.GOOGLE_DRIVE_TOKEN}" ]]; then
+                            echo "File exists. Deleting it..."
+                            rm -f "${env.GOOGLE_DRIVE_TOKEN}"
+                        fi
+
+                        echo "Copying secret file to the temporary local file..."
+                        cp "$GOOGLE_DRIVE_CREDENTIALS_FILE" "${env.GOOGLE_DRIVE_CREDENTIALS}"
+                        cp "$GOOGLE_DRIVE_TOKEN_FILE" "${env.GOOGLE_DRIVE_TOKEN}"
+                        """
+                    }
             }
         }
 
@@ -197,8 +227,8 @@ END
                                         -e IBM_BUCKET='${IBM_BUCKET}' \
                                         -e IBM_KEY='${IBM_KEY}' \
                                         -e GOOGLE_DRIVE_PATH='${GOOGLE_DRIVE_PATH}' \
-                                        -e GOOGLE_DRIVE_CREDENTIALS_FILE='${GOOGLE_DRIVE_CREDENTIALS_FILE}' \
-                                        -e GOOGLE_DRIVE_TOKEN_FILE='${GOOGLE_DRIVE_TOKEN_FILE}' \
+                                        -e GOOGLE_DRIVE_CREDENTIALS_FILE='${DESTINATION_PATH}/${CREDENTIALS_FILE}' \
+                                        -e GOOGLE_DRIVE_TOKEN_FILE='${DESTINATION_PATH}/${TOKEN_FILE}' \
                                         -e GOOGLE_DRIVE_SHARED_DRIVE_ID='${GOOGLE_DRIVE_SHARED_DRIVE_ID}' \
                                         -e RUN_ARTIFACTS_URL='${RUN_ARTIFACTS_URL}' \
                                         -e BUILD_VERSION='${build_version}' \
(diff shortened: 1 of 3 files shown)


jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile
```diff
@@ -109,6 +109,33 @@ END
...
+        stage('Prepare Google credential files') {
+            steps {
+                    script {
+                        // Assign the file path to the environment variable
+                        env.GOOGLE_DRIVE_CREDENTIALS = "${env.WORKSPACE}/credentials.json"
+                        env.GOOGLE_DRIVE_TOKEN = "${env.WORKSPACE}/token.json"
+
+                        sh """
+                        echo "Checking if ${env.GOOGLE_DRIVE_CREDENTIALS} exists..."
+                        if [ -f "${env.GOOGLE_DRIVE_CREDENTIALS}" ]; then
```
RobertKrawitz, line 121: Use `if [[ ... ]]` syntax; single bracket is deprecated

jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile
```diff
@@ -109,6 +109,33 @@ END
...
+                            echo "File exists. Deleting it..."
+                            rm -f "${env.GOOGLE_DRIVE_CREDENTIALS}"
+                        fi
+                        echo "Checking if ${env.GOOGLE_DRIVE_TOKEN} exists..."
+                        if [ -f "${env.GOOGLE_DRIVE_TOKEN}" ]; then
```
RobertKrawitz, line 126: Same, use double square bracket syntax.

jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile
```diff
@@ -109,6 +109,33 @@ END
...
+                            echo "File exists. Deleting it..."
+                            rm -f "${env.GOOGLE_DRIVE_TOKEN}"
+                        fi
+
+                        echo "Copying secret file to the temporary local file..."
+                        cp "$GOOGLE_DRIVE_CREDENTIALS_FILE" "${env.GOOGLE_DRIVE_CREDENTIALS}"
+                        cp "$GOOGLE_DRIVE_TOKEN_FILE" "${env.GOOGLE_DRIVE_TOKEN}"
```
RobertKrawitz, line 133: These should not be world readable on general principles; either make sure that the directory containing them is mode 700 or use `umask 77` to make the files themselves unreadable to anyone other than owner.

jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile
```diff
@@ -197,8 +224,8 @@ END
                                         -e IBM_BUCKET='${IBM_BUCKET}' \
                                         -e IBM_KEY='${IBM_KEY}' \
                                         -e GOOGLE_DRIVE_PATH='${GOOGLE_DRIVE_PATH}' \
-                                        -e GOOGLE_DRIVE_CREDENTIALS_FILE='${GOOGLE_DRIVE_CREDENTIALS_FILE}' \
-                                        -e GOOGLE_DRIVE_TOKEN_FILE='${GOOGLE_DRIVE_TOKEN_FILE}' \
+                                        -e GOOGLE_DRIVE_CREDENTIALS_FILE='/tmp/credentials.json' \
+                                        -e GOOGLE_DRIVE_TOKEN_FILE='/tmp/token.json' \
```
RobertKrawitz, line 228: Where do these filenames  come from?  In any case, use mktemp-generated temporaries rather than fixed filenames which could be subject to various attacks.

RobertKrawitz (changes requested):
Same comments apply to all three jenkinsfiles.

jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile
ebattat: As I said before it is internal Jenkins under VPN

jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile
RobertKrawitz: It's good practice to always handle this correctly.  Other users of this might not be on a VPN and might need this protection.

jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile
ebattat: Those file come from secrets and cannot be shared with anyone:
https://github.com/redhat-performance/benchmark-runner/blob/5e1a0e52905bf51e28227f168eb1d3499707e042/jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile#L23-L24

RobertKrawitz: But it looks like you're copying them into local files in /tmp?

ebattat: Yes, this is the only way that I found that Jenkins works, otherwise it failed to read it.

ebattat: I will put it in secret

ebattat: Pls take a look now, its fully secured

jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile
RobertKrawitz: I still think you should `umask 77` here.  It's a trivial change but worth doing correctly.

jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile
```diff
@@ -197,8 +227,8 @@ END
                                         -e IBM_BUCKET='${IBM_BUCKET}' \
                                         -e IBM_KEY='${IBM_KEY}' \
                                         -e GOOGLE_DRIVE_PATH='${GOOGLE_DRIVE_PATH}' \
-                                        -e GOOGLE_DRIVE_CREDENTIALS_FILE='${GOOGLE_DRIVE_CREDENTIALS_FILE}' \
-                                        -e GOOGLE_DRIVE_TOKEN_FILE='${GOOGLE_DRIVE_TOKEN_FILE}' \
+                                        -e GOOGLE_DRIVE_CREDENTIALS_FILE='${DESTINATION_PATH}/${CREDENTIALS_FILE}' \
+                                        -e GOOGLE_DRIVE_TOKEN_FILE='${DESTINATION_PATH}/${TOKEN_FILE}' \
```
RobertKrawitz, line 231: Are DESTINATION_PATH and CREDENTIALS_FILE constant, or do they change from run to run (so that if e. g. you're doing two concurrent runs they would not collide)?

RobertKrawitz (changes requested):

jenkins/PerfCI_Chaos/03_PerfCI_Choas_Windows_VMs_Deployment/Jenkinsfile
ebattat: Every run those files are deleted at the end.

RobertKrawitz: 1. If they're fixed file names, someone could guess them.
2. What happens if there are multiple concurrent runs?

They really should be distinct random filenames, or at least within distinct randomly generated directory names.

RobertKrawitz: Non-colliding directory names that are mode 700 would be fine.

ebattat: Sorry Robert, I am not going to spend more time on Jenkins.

RobertKrawitz: OK, then I'll open an issue to record it and approve this.

RobertKrawitz (approved):

PR #956, opened 2025-01-13
Update get ODF version

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

The ODF version metadata was changed in ODF 4.17. 
I have keep the same method for OCP <= 4.16 and update it for greater.


RobertKrawitz (approved):
Not the cleanest way to do it, but should work well enough.

PR #960, opened 2025-01-23
OCP4.17: self-node-remediation-operator

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

self-node-remediation-operator should be installed in OCP4.17,
for more details [see](https://docs.redhat.com/en/documentation/workload_availability_for_red_hat_openshift/25.1/html/release_notes/workload_availability_for_red_hat_openshift_25_1_release_notes#twenty-five.1-deprecated-removed)


benchmark_runner/common/ocp_resources/create_nhc_snr.py
```diff
@@ -26,16 +26,15 @@ def create_nhc_snr(self):
             logger.info(f'run {resource}')
             self.__oc.create_async(yaml=os.path.join(self.__path, resource))
 
-            if '03_subscription.yaml' in resource:
-                # verify once after create all resource files
+            if '03_nhc_subscription.yaml' in resource:
```
RobertKrawitz, line 29: For the future, we should not have to hard code all of this inline in this way.  But that's a separate issue and not a problem with this PR per se.

RobertKrawitz (approved):

PR #974, opened 2025-02-09
Hot fix: add API server to login

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description:
encountered the following error when running 'oc login' inside a container:
`subprocess.TimeoutExpired: Command 'oc login -u kubeadmin -p password' timed out after 8000 seconds
`
Adding the API server resolved the issue:
`oc login https://api.<cluster-name>.<domain>:6443 -u <USERNAME> -p <PASSWORD>`

A kubeconfig file is required to extract the API server.


benchmark_runner/common/oc/oc.py
```diff
@@ -715,6 +717,22 @@ def collect_events(self):
         self.run(f"{self.__cli} get events -A > '{output_filename}' ")
         return output_filename
 
+    def get_kube_api_server(self):
+        """
+        This method returns server url from config file
+        :return: server url
+        """
+        try:
+            with open(self.__kubeconfig_path, "r") as f:
```
RobertKrawitz, line 726: The kubeconfig file is a YAML file.  Rather than reading the file in this way, load the YAML file appropriately and extract the correct entry.

RobertKrawitz (changes requested):
Use the yaml module rather than reading the kubeconfig file yourself.

benchmark_runner/common/oc/oc.py
ebattat: @RobertKrawitz  Thx for feedback,
I have updated it using yaml package.

benchmark_runner/common/oc/oc.py
```diff
@@ -715,6 +717,20 @@ def collect_events(self):
         self.run(f"{self.__cli} get events -A > '{output_filename}' ")
         return output_filename
 
+    def get_kube_api_server(self):
+        try:
+            with open(self.__kubeconfig_path, "r") as f:
+                config = yaml.safe_load(f)
+
+            # Extract the API server URL from the clusters section
+            return config["clusters"][0]["cluster"]["server"]
```
RobertKrawitz, line 726: This is probably going to be OK in practice since you have control over the kubeconfig file and there isn't likely to be more than one cluster in the file, but you may want to consider using name of the cluster for lookup.  I'm going to approve it, but it may be a problem down the road at some point.

RobertKrawitz (approved):

PR #978, opened 2025-02-12
Hotfix: remove readonly and use root

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description
Remove the read-only config file and use the root config instead of the Jenkins config, as the script runs with sudo.
read-only cause to the following error: 
`subprocess Status : FAIL: 1 b'Login successful.\n\nYou have access to 71 projects, the list has been suppressed. You can list all projects with \'oc projects\'\n\nUsing project "default".\nerror: open /****/.kube/config: read-only file system\n'`


RobertKrawitz (approved):
Please make sure to test this before merging; it's unfortuate to have to revert this fix.

PR #984, opened 2025-02-19
Class-Level Singleton for login

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Currently, the login method can be called multiple times per workload. Adding a Class-Level Singleton for login will solve this issue.


RobertKrawitz (commented):
What I suggest you consider is for oc.login() to be a private member that's invoked by other oc members as needed, and then to remove explicit oc.login() calls from the rest of the code.

ebattat:
I dont think login should be a private method, oc.login is created once per workload.

ebattat:
Thats what happened today.
It is done in workload class [initialization](https://github.com/redhat-performance/benchmark-runner/blob/2d4a340710a0cfaf3491039348df628ec6a37b73/benchmark_runner/workloads/workloads_operations.py#L93-L95)

RobertKrawitz:
I'm suggesting not doing it at class initialization, but only when an oc call is made requiring login.  Or at least for when the oc class is initialized, not each subclass.  There are quite a few calls to various login() methods scattered about benchmark_runner.

RobertKrawitz:
Understood; what I'm suggesting is to have the code that calls oc login be in one place (inside oc.py, so nothing else needs to know about oc login).

ebattat:
Thanks, Robert, for this important comment.
I have changed the login to a private method and only call it inside OC class once, canceling all other login calls outside the OC class.

RobertKrawitz (approved):

RobertKrawitz:
This is a lot cleaner, approved.

PR #985, opened 2025-02-20
Remove login server api url

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Remove login server api url, extra parameter that is not required for oc login command.


RobertKrawitz (commented):
Why don't you need the kubeconfig path?

ebattat:
I have removed the API server from the 'oc login' command, so there is no need to extract it from the kubeconfig anymore

RobertKrawitz:
How does oc login know what to login to without the API server?

ebattat:
We mount the kube config file path

RobertKrawitz:
And then the KUBECONFIG environment variable is set correctly?

ebattat:
Yes,
https://github.com/redhat-performance/benchmark-runner/blob/54dcadf9746a19ad99d5e48598dacbabf74965d7/jenkins/PerfCI/03_PerfCI_Workloads_Deployment/Jenkinsfile#L194

RobertKrawitz:
Got it.  So it centralizes KUBECONFIG handling in one place (the jenkinsfile) and nothing else then has to know about it?

ebattat:
Right

RobertKrawitz (approved):
Good idea encapsulating one more thing.

I wonder whether you could do the same kind of thing with KUBEADMIN_PASSWORD.

ebattat:
KUBEADMIN_PASSWORD is a string and not a file, so I don’t think it’s the right thing to mount.

PR #990, opened 2025-02-25
Secret OCP and Operator Versions

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. Add OCP version secret
2. Add OCP Client secret
3. Add CNV/ODF/LSO operator secrets
4. Pass OCP client version into Dockerfile as argument

### DIFF
--- a/.github/workflows/Perf_Env_Build_Test_CI.yml
+++ b/.github/workflows/Perf_Env_Build_Test_CI.yml
@@ -255,8 +255,8 @@ jobs:
         run: |
           pip install setuptools
           version=$(python3 setup.py --version)
-          sudo docker build --build-arg VERSION=$version -t ${{ secrets.QUAY_PRIVATE_BENCHMARK_RUNNER_REPOSITORY }}:v$version .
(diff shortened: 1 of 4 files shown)


.github/workflows/Perf_Env_Build_Test_CI.yml
```diff
@@ -255,10 +255,8 @@ jobs:
         run: |
           pip install setuptools
           version=$(python3 setup.py --version)
-          sudo docker build --build-arg VERSION=$version -t ${{ secrets.QUAY_PRIVATE_BENCHMARK_RUNNER_REPOSITORY }}:v$version .
```
RobertKrawitz, line 258: There should be a versioned name as well as latest, so it's easy to back off to an earlier version if needed.

RobertKrawitz (commented):
My only issue is with removing the versioned images, otherwise looks great.

.github/workflows/Perf_Env_Build_Test_CI.yml
ebattat: Agree, keep it

RobertKrawitz (approved):

PR #992, opened 2025-03-02
Add choas_ci run type

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Add chaos_ci run type for dedicated Elasticsearch index.


benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -392,8 +392,8 @@ def run(self):
         """
         try:
             self._initialize_run()
-            if self._run_type == 'test_ci':
-                self._es_index = 'bootstorm-test-ci-results'
+            if self._run_type in ('test_ci', 'chaos_ci'):
+                self._es_index = f"bootstorm-{self._run_type.replace('_', '-')}-results"
```
RobertKrawitz, line 396: Why are these cases hard coded?

ebattat: Its the dynamic elasticsearch index name
What do u mean hard coded ?

RobertKrawitz: Names of explicit run types (in this case) special cased in the code.  But why are you using `chaos-ci` rather than `chaos_ci` as the index name?  We're  not doing that with func_ci and perf_ci.

I think I remember now that this came from the elasticsearch index name.  Please add a comment

ebattat: Done.

RobertKrawitz (approved):

PR #1000, opened 2025-03-13
Fix missing quotes

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Fix missing quotes

### DIFF
--- a/jenkins/PerfCI_Chaos/04_PerfCI_Chaos_Upgrade_OpenShift_Deployment/Jenkinsfile
+++ b/jenkins/PerfCI_Chaos/04_PerfCI_Chaos_Upgrade_OpenShift_Deployment/Jenkinsfile
@@ -151,8 +151,8 @@ END
                                                         sudo podman run --rm \
                                                             -e UPGRADE_OCP_VERSION="${UPGRADE_OCP_VERSION}" \
                                                             -e UPGRADE_STEP="${upgrade_step}" \
-                                                            -e RUN_TYPE="${RUN_TYPE} \
-                                                            -e UPGRADE_CHANNEL='${UPGRADE_CHANNEL}' \
+                                                            -e RUN_TYPE="${RUN_TYPE}" \
+                                                            -e UPGRADE_CHANNEL="${UPGRADE_CHANNEL}" \
                                                             -e LSO_VERSION="${LSO_VERSION}" \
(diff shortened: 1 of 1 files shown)


jenkins/PerfCI_Chaos/04_PerfCI_Chaos_Upgrade_OpenShift_Deployment/Jenkinsfile
```diff
@@ -151,7 +151,7 @@ END
                                                         sudo podman run --rm \
                                                             -e UPGRADE_OCP_VERSION="${UPGRADE_OCP_VERSION}" \
                                                             -e UPGRADE_STEP="${upgrade_step}" \
-                                                            -e RUN_TYPE="${RUN_TYPE} \
+                                                            -e RUN_TYPE="${RUN_TYPE}" \
```
RobertKrawitz, line 154: You also need to fix UPGRADE_CHANNEL and EXPECTED_NODES, both of which use single quotes rather than double.

If it's working without those fixes, then those variables aren't actually needed here.

RobertKrawitz (changes requested):

jenkins/PerfCI_Chaos/04_PerfCI_Chaos_Upgrade_OpenShift_Deployment/Jenkinsfile
ebattat: Fixed only UPGRADE_CHANNEL, EXPECTED_NODES is a dictionary which sensitive to double quotes in groovy

RobertKrawitz: So the problem is that the `podman run` command will get literally on the command line (unless something else expands it):

```
podman run -e EXPECTED_NODES=${EXPECTED_NODES}
```

with no expansion of EXPECTED_NODES.

ebattat: The problem is when run it through Jenkinsfile (groovy) and not in command line

RobertKrawitz (approved):

PR #1002, opened 2025-03-26
ODF storage cleanup

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Adding local-storage cleanup before LSO/ ODF installation


benchmark_runner/common/ocp_resources/create_odf.py
```diff
@@ -42,7 +42,7 @@ def create_odf(self, upgrade_version: str):
                         for node, disk_ids in self.__worker_disk_ids.items():
                             for disk_id in disk_ids:
                                 disk = f'/dev/disk/by-id/{self.__worker_disk_prefix}{disk_id}'
-                                delete_node_disk += f"""sudo sgdisk --zap-all {disk}; sudo wipefs -a {disk}; sudo dd if=/dev/zero of='{disk}' bs=1M count=100 oflag=direct,dsync; sudo blkdiscard {disk}; sudo partprobe {disk};"""
+                                delete_node_disk += f"""sudo sgdisk --zap-all {disk}; sudo wipefs -a {disk}; sudo dd if=/dev/zero of='{disk}' bs=10M count=100 oflag=direct,dsync; sudo blkdiscard {disk}; sudo partprobe {disk};"""
```
RobertKrawitz, line 45: What's the reason for this?

benchmark_runner/common/ocp_resources/odf/template/01_delete_disks_template.sh
```diff
@@ -4,5 +4,6 @@ while (($# >= 2)) ; do
     node="$1"
     delete_disk_command="$2"
     oc debug "node/$node"  -- chroot /host sh -c "$delete_disk_command"
+    oc debug "node/$node" -- chroot /host sh -c "[ -d /mnt/local-storage/ ] && rm -rf /mnt/local-storage/ || echo 'Directory /mnt/local-storage/ does not exist'"
```
RobertKrawitz, line 7: Where does /mnt/local_storage come from?  Could it be passed in to the script as a third argument?

ebattat: Its hard coded path for ODF storage

benchmark_runner/common/ocp_resources/create_odf.py
gqlo: I have done some investigation and wiping out the first 1GB of disk might not be sufficent. I found the problmatic disk and ODF failed to provision the disk because it detected some meta data at offset slightly over 1GiB position, this makes me think they are scaning the whole disk. for now, the workaround can be just wiping the entire disk. 
  
```
sh-5.1# strings -t d /dev/sdf | grep -A5 -B5 "8e10ea03"
1073742017 4096
1073742025 bfm_size
1073742037 959656755200
1073742053 bluefs
1073742068 ceph_fsid$
1073742081 8e10ea03-6275-4002-9a32-c136cbb3914e
1073742121 ceph_version_when_createdV
1073742150 ceph version 19.2.0-98.el9cp (d5c3cf625491b0bd76b4585e77aa0d907446f314) squid (stable)
1073742240 created_at
1073742254 2025-03-23T10:58:43.010337Z
1073742285 elastic_shared_blobs
```

gqlo (approved):
only as a workaround for now, we need to understand how ceph store its metadata especially when large disks are used.

benchmark_runner/common/ocp_resources/odf/template/01_delete_disks_template.sh
gqlo: This is not need if you are doing a complete fresh install.

ebattat: Removed

benchmark_runner/common/ocp_resources/create_odf.py
RobertKrawitz: Is blkdiscard not doing the job or is it failing?  It should clear everything out by calling TRIM on the entire disk.

ODF likely stores superblock type information in multiple places on the disk, but unfortunately I don't see any documentation on the format or instructions for how to prepare previously used disks.

gqlo: both blkdiscard and partprobe are not working, the deletion code is too simple at the moment, it seems like it doesn't report any error when it fails. 

```
blkdiscard: /dev/disk/by-id/wwn-0x62cea7f08df680002d419bea0ee29b52: BLKDISCARD ioctl failed: Operation not supported
sudo: partprobe: command not found
```

ebattat: I have removed both cmd that are not working:
`sudo blkdiscard {disk}; sudo partprobe {disk};`

benchmark_runner/common/ocp_resources/create_odf.py
```diff
@@ -42,7 +42,7 @@ def create_odf(self, upgrade_version: str):
                         for node, disk_ids in self.__worker_disk_ids.items():
                             for disk_id in disk_ids:
                                 disk = f'/dev/disk/by-id/{self.__worker_disk_prefix}{disk_id}'
-                                delete_node_disk += f"""sudo sgdisk --zap-all {disk}; sudo wipefs -a {disk}; sudo dd if=/dev/zero of='{disk}' bs=1M count=100 oflag=direct,dsync; sudo blkdiscard {disk}; sudo partprobe {disk};"""
+                                delete_node_disk += f"""sudo sgdisk --zap-all {disk}; sudo wipefs -a {disk}; sudo dd if=/dev/zero of='{disk}' bs=1M count=100 oflag=direct,dsync; """
```
gqlo, line 45: I was thinking about something like this. so we can check if any of the command executed succesfully or not and which disk it failed operating on. 

`sudo sgdisk --zap-all {disk} && echo "✓ GPT/MBR zapped on {disk}" || echo "✗ GPT/MBR zap failed on {disk}"; sudo wipefs -a {disk} && echo "✓ Signatures wiped on {disk}" || echo "✗ Signature wipe failed on {disk}"; sudo dd if=/dev/zero of='{disk}' bs=1M count=100 oflag=direct,dsync && echo "✓ Zeroing successful on {disk}" || echo "✗ Zeroing failed on {disk}"`

gqlo (changes requested):

benchmark_runner/common/ocp_resources/create_odf.py
ebattat: No need to add echo inside the code. I will print the output of the whole shell, so if there is any error, we will see it while running.

benchmark_runner/common/ocp_resources/create_odf.py
RobertKrawitz: You should investigate why blkdiscard is not working.  Start by running it against a drive with some known (but not important!) data on it to see what happens by looking for data on it afterwards.

It's possible that the drive does not support TRIM, but I'm pretty sure all modern SSDs do.  It's also possible that the drive itself is virtualized , in which case the virtualization layer may not pass it through.

gqlo: Those disks might be HDDs. 
```
sh-5.1# lsblk -d -o name,rota,disc-max,disc-zero | grep sd*
sda        1       0B         0
sdb        1       0B         0
sdc        1       0B         0
sdd        1       0B         0
sde        1       0B         0
sdf        1       0B         0
sdg        1       0B         0
```

PR #1003, opened 2025-03-28
Update number of ODF disks

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

There are 7 disk per worker node, thought to update it to 7 instread of 6
I am afraid that it will change the benchmark numbers, need to double check if we can merge it.


ebattat:
6 related to ODF and 1 related to LSO, thats the reason we choose 6 for ODF

PR #1008, opened 2025-04-06
bootstorm VMs parallel verification

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. Added parallel verification for bootstorm VMs: Previously, the verification was done in a serial loop. By switching to parallel verification, the verification time has been significantly improved.

2.verify_after_test variable: This variable is used in the catalog to indicate when the verification has been completed.

### DIFF
--- a/benchmark_runner/workloads/bootstorm_vm.py
+++ b/benchmark_runner/workloads/bootstorm_vm.py
@@ -1,7 +1,7 @@
 
 import os
 import time
-from multiprocessing import Process
+from multiprocessing import Process, Manager
 
 from benchmark_runner.common.logger.logger_time_stamp import logger_time_stamp, logger
 from benchmark_runner.common.elasticsearch.elasticsearch_exceptions import ElasticSearchDataNotUploaded
@@ -162,7 +162,7 @@ class BootstormVM(WorkloadsOperations):
             try:
                 virtctl_status = self._oc.get_virtctl_vm_status(vm_name)
 
-                if virtctl_status == "True":
+                if str(virtctl_status).lower() == "true":
                     # Log success and store relevant data
                     logger.info(f"VM {vm_name} verified successfully (virtctl SSH status: { virtctl_status}).")
                     break  # Exit loop on success
@@ -173,22 +173,25 @@ class BootstormVM(WorkloadsOperations):
                 logger.info(f"Attempt {attempt + 1}/{retries} failed for Virtctl SSH VM {vm_name}: {e}")
 
             # Sleep before retrying
-            time.sleep(delay)
+            if attempt < retries - 1:
+                time.sleep(delay)
+
 
         # Final update to self._data_dict after all attempts
         vm_node = self._oc.get_vm_node(vm_name)  # Get the node again in case it changed
         self._data_dict = {
             'vm_name': vm_name,
             'node': vm_node,
-            'virtctl_vm': 1 if virtctl_status == 'True' else 0, # int value for Grafana
+            'virtctl_vm': 1 if str(virtctl_status).lower() == "true" else 0, # int value for Grafana
             'virtctl_status': virtctl_status,
+            'verify_after_test': self._verify_after_test,
             'run_artifacts_url': os.path.join(
                 self._run_artifacts_url,
                 f"{self._get_run_artifacts_hierarchy(self._workload_name, True)}-{self._time_stamp_format}.tar.gz"
             )
         }
 
-        if virtctl_status != "True":
+        if str(virtctl_status).lower() != "true":
             logger.info(
                 f"All attempts failed for VM {vm_name}. Final SSH status: {self._data_dict.get('virtctl_status', 'No status available')}")
             error_log_path = f"{self._run_artifacts_path}/{vm_name}_error.log"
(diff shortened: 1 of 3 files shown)


benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -173,99 +173,133 @@ def _verify_single_vm(self, vm_name, retries=5, delay=10):
...
+                p = Process(target=self._verify_single_vm_wrapper, args=(vm_name, return_dict))
+                p.start()
+                processes.append(p)
+
+            for p in processes:
+                p.join()
+
+        # After all processes are done, collect failures
+        for vm_name, status in return_dict.items():
+            if status != 'True':
```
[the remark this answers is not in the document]
ebattat: Fixed

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -173,99 +173,132 @@ def _verify_single_vm(self, vm_name, retries=5, delay=10):
...
+                p = Process(target=self._verify_single_vm_wrapper, args=(vm_name, return_dict))
+                p.start()
+                processes.append(p)
+
+            for p in processes:
+                p.join()
+
+        # After all processes are done, collect failures
+        for vm_name, status in return_dict.items():
+            if status.lower() != 'true':
```
RobertKrawitz, line 246: Does this need to be
```
if str(status).lower() != 'true':
```
If not, do the other ones above need that?

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -173,99 +173,132 @@ def _verify_single_vm(self, vm_name, retries=5, delay=10):
...
+                    upgrade_done = self._oc.get_cluster_status() == f'Cluster version is {self._wait_for_upgrade_version}'
+                    current_wait_time = 0
+
+                    while (self._timeout <= 0 or current_wait_time <= self._timeout) and not upgrade_done:
+                        failure_vms = self._verify_vms_in_parallel(vm_names)
+                        upgrade_done = self._oc.get_cluster_status() == f'Cluster version is {self._wait_for_upgrade_version}'
+
+                        # Sleep 1 sec between each cycle
+                        time.sleep(1)
+                        current_wait_time += 1  # Increment the wait time
```
RobertKrawitz, line 274: The verification probably takes a fair bit of time, so rather than assuming that the 1 second delay is the actual wait time,  get the starting time before starting the loop and compare the current time against the starting time for checking timeout.

Also, given that this is probably fairly resource-intensive, a 1 second interval between checks might be a bit too short and put unnecessary load on the server.

RobertKrawitz: That's still going to add up fast and if you're verifying 240 at once that will be a lot of load on the server.

ebattat: Ok, I’ve updated it to use a constant SLEEP_TIME of 3 seconds instead of 1.
BTW, we have a verification mechanism that handles each VM individually. 
I added this delay just to be on the safe side.

ebattat:
@RobertKrawitz, can we merge it ?

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -173,99 +173,132 @@ def _verify_single_vm(self, vm_name, retries=5, delay=10):
...
+                    upgrade_done = self._oc.get_cluster_status() == f'Cluster version is {self._wait_for_upgrade_version}'
+                    current_wait_time = 0
+
+                    while (self._timeout <= 0 or current_wait_time <= self._timeout) and not upgrade_done:
+                        failure_vms = self._verify_vms_in_parallel(vm_names)
+                        upgrade_done = self._oc.get_cluster_status() == f'Cluster version is {self._wait_for_upgrade_version}'
+
+                        # Sleep 3 sec between each cycle
+                        time.sleep(self.SLEEP_TIME)
+                        current_wait_time += 1  # Increment the wait time
```
RobertKrawitz, line 274: I'm still concerned about the timeout/server load issue I raised the other day.  It appears to me that you're verifying all of the VMs in parallel and then sleeping for a second.  There are two problems here:

1. You're almost certainly going to find that the actual timeout will be a lot longer than the desired timeout, because the verification step will take much longer than a second whether you're parallelizing them or not.  The right way to do this is to get the timestamp prior to entering the loop and then checking the current time against the timestamp rather than simply incrementing the wait time.
2. Sleeping only 1 second between each mass check will result in a lot of API traffic.

RobertKrawitz: Still creates a fair bit of traffic, and the timeout is going to be wrong.  I suggest checking every 5 seconds unless you have a good reason otherwise, and comparing time.time() against the start rather than counting iterations.

ebattat: Agree, I have updated it to 10 sec.

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -173,99 +173,133 @@ def _verify_single_vm(self, vm_name, retries=5, delay=10):
...
+                    upgrade_done = self._oc.get_cluster_status() == f'Cluster version is {self._wait_for_upgrade_version}'
+                    current_wait_time = 0
+
+                    while (self._timeout <= 0 or current_wait_time <= self._timeout) and not upgrade_done:
+                        failure_vms = self._verify_vms_in_parallel(vm_names)
+                        upgrade_done = self._oc.get_cluster_status() == f'Cluster version is {self._wait_for_upgrade_version}'
+
+                        # Sleep between each cycle
+                        time.sleep(delay)
+                        current_wait_time += delay  # Increment the wait time
```
RobertKrawitz, line 275: It should be something like this:
```
                    update_start_time = time.time()
                    while (self._timeout <= 0 or time.time() - update_start_time <= self._timeout) and not upgrade_done:
                        failure_vms = self._verify_vms_in_parallel(vm_names)
                        upgrade_done = self._oc.get_cluster_status() == f'Cluster version is {self._wait_for_upgrade_version}'

                        # Sleep between each cycle
                        time.sleep(delay)
```

RobertKrawitz (changes requested):
The timeout code is still not correct.

benchmark_runner/workloads/bootstorm_vm.py
ebattat: No need I have a timeout counter (up to 3600 sec), the start time is not relevant for timeout.

RobertKrawitz: You don't know how long the check takes.  What happens if it takes 20 seconds each time?  A minute?  Then the "3600 second" timeout is actually 10800 seconds.  Rather than assume that the current_wait_time actually increments by the delay, just compare the current time against the starting time.

ebattat: This is the way I have handled timeout in benchmark-runner.
`current_wait_time += delay `  it will failed when reaching to 3600, its default timeout.

RobertKrawitz: But you don't know what the *actual* delay is.  What if _verify_vms_in_parallel were to take 20 seconds, for example?

ebattat: Agree, Fixed

ebattat: Agree, Fixed

RobertKrawitz (approved):

PR #1009, opened 2025-04-06
fix krknhub kubeconfig

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Fix krknhub kubeconfig

### DIFF
--- a/benchmark_runner/krkn_hub/krknhub_workloads.py
+++ b/benchmark_runner/krkn_hub/krknhub_workloads.py
@@ -15,6 +15,7 @@ class KrknHubWorkloads(WorkloadsOperations):
         # environment variables
         self.__namespace = self._environment_variables_dict.get('namespace', '')
         self.__krknhub_workload = self._environment_variables_dict.get('krknhub_workload', '')
+        self.__provision_kubeconfig_path = self._environment_variables_dict.get('provision_kubeconfig_path', '')
         self.__krknhub_environment_variables = self._environment_variables_dict.get('krknhub_environment_variables', '')
         self.__ssh = SSH()
         self.__krknhub_pod_name = 'bm'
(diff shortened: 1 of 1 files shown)


ebattat:
Regarding the Copilot comment, there's no need to check for the existence of this file because it will already fail at an earlier step if the file doesn't exist.

benchmark_runner/krkn_hub/krknhub_workloads.py
```diff
@@ -53,7 +54,7 @@ def run_workload(self):
         """
         logger.info(f'run krkn-hub: {self.__krknhub_workload}')
         pod_log_path = f'{self._run_artifacts_path}/{self.__krknhub_workload}_pod.log'
-        workload_command = f'{self.__krknhub_environment_variables}; podman run --rm --name={self.__krknhub_pod_name} --net=host --env-host=true -v /root/.kube/config:/home/krkn/.kube/config:Z quay.io/krkn-chaos/krkn-hub:{self.__krknhub_workload} > {pod_log_path}'
+        workload_command = f'{self.__krknhub_environment_variables}; chmod 775 {self.__provision_kubeconfig_path} ; podman run --rm --name={self.__krknhub_pod_name} --net=host --env-host=true -v {self.__provision_kubeconfig_path}:/home/krkn/.kube/config:Z quay.io/krkn-chaos/krkn-hub:{self.__krknhub_workload} > {pod_log_path}'
```
RobertKrawitz, line 57: Use the most restrictive permissions you can for this, which is probably 600.  The kubeconfig file isn't executable, so should not have the executable bit set; also, it almost certainly does not need group write and global read permission.

ebattat: `chmod 600` didnt work, change it to `chmod 644`

RobertKrawitz: OK, I guess something else needs to be able to access the file.

RobertKrawitz (approved):

PR #1015, opened 2025-04-11
Scale Windows or Fedora VMs

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

This document provides automated scripts and configuration to **scale Fedora and Windows VMs** on OpenShift clusters. It verifies the VM login page and measures **bootstorm time**—the time it takes for VMs to become fully operational.
Add option to configure fedora container disk image .qcow2 image from quay.io (default: Fedora 37)
and also Windows virtio qcow2 image.

### DIFF
--- a/docs/source/scaleVM.md
+++ b/docs/source/scaleVM.md
@@ -1,0 +1,210 @@
+# 🖥️ Scale Fedora & Windows VMs with Benchmark Runner
+
+This document provides automated scripts and configuration to **scale Fedora and Windows VMs** on OpenShift clusters. It verifies the VM login page and measures **bootstorm time**—the time it takes for VMs to become fully operational.
+
+---
+
+## 🚀 Features
+
+- Scale Fedora VMs using **ephemeral storage**
+- Scale Windows VMs using **ODF-backed storage** (via DataVolume)
+- Parallel VM loading across specified worker nodes
+- Optional metrics collection via **Elasticsearch**
+- Logs and artifacts are stored locally for post-run analysis
+
+---
+
+## 🧪 Prerequisites
+
+- OpenShift cluster with kubeadmin access
+- [`podman`](https://podman.io) installed
+
+---
+
+## ⚙️ Parameters
+
+### Common Parameters
+| Variable                    | Description                                                                                                                                                                                                                                                                 |
+|-----------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `KUBEADMIN_PASSWORD`        | *(Optional, if already login)* OpenShift admin password                                                                                                                                                                                                                     |
+| `KUBECONFIG_PATH`           | Path to your kubeconfig (e.g., `$HOME/.kube/config`)                                                                                                                                                                                                                        |
+| `SCALE_NODES`               | List of target worker nodes (e.g., `["worker-0", "worker-1"]`)                                                                                                                                                                                                              |
+| `THREADS_LIMIT`             | Number of physical CPUs to use for parallelism (e.g., `20`)                                                                                                                                                                                                                 |
+| `TIMEOUT`                   | Maximum duration (in seconds) for the run; `0` or a negative value means no timeout                                                                                                                                                                                         |
+| `LOGS_PATH`                 | *(Optional)* Path for logs (YAML files and `benchmark_runner.log`); default is `/tmp/benchmark-runner-run-artifacts`                                                                                                                                                        |
+| `RUN_ARTIFACTS_URL`         | *(Optional, default: `False`)* If `True`, Remote URL for saving logs                                                                                                                                                                                                        |
+| `SAVE_ARTIFACTS_LOCAL`      | *(Optional, default: `False`)* If `True`, saves logs locally to `/tmp/benchmark-runner-run-artifacts`                                                                                                                                                                       |
+| `ENABLE_PROMETHEUS_SNAPSHOT` | *(Optional, default: `False`)* If `True`, saves and collects a Prometheus snapshot (`promdb_*.tar`) and query results. See [metrics-default.yaml](https://github.com/redhat-performance/benchmark-runner/blob/main/benchmark_runner/common/prometheus/metrics-default.yaml) |
+| `RUN_STRATEGY` | *(Optional, default: `False`)* If `True`, when `RUN_STRATEGY` is defined, it injects `runStrategy: Always` into the YAML.                                                                                                                                                   |
+
+### Results in `benchmark_runner.log` (can be distributed into ElasticSearch):
+- `vm_name`: Name of the VM  
+- `node`: The node the VM ran on  
+- `bootstorm_time` (milliseconds): Time taken to boot the VM  
+- `vm_ssh`: `1` = success, `0` = failure; indicates whether the VM is reachable via SSH  
+
+## 🧰 Fedora VM BootStorm (Ephemeral Storage)
+
+This mode loads Fedora VMs using ephemeral storage and verifies their readiness via the login page.
+
+### Fedora-Specific Parameters
+
+| Variable | Description                                                                                      |
+|----------|--------------------------------------------------------------------------------------------------|
+| `BOOTSTORM_SCALE` | Number of Fedora VMs per worker node, e.g. 10                                                    |
+| `FEDORA_CONTAINER_DISK` | (Optional) Fedora .qcow2 image from quay.io (e.g.: quay.io/openshift-cnv/qe-cnv-tests-fedora:37) |
+
+🔧 Example Configuration:
+```bash
+# OpenShift admin password
+export KUBEADMIN_PASSWORD=xxxxx-xxxxx-xxxxx-xxxxx
+# Number of Windows VMs per worker node
+export BOOTSTORM_SCALE=10
+# container disk image
+export FEDORA_CONTAINER_DISK="quay.io/openshift-cnv/qe-cnv-tests-fedora:37"
+# List of target worker nodes
+export SCALE_NODES="[ 'worker-0', 'worker-1', 'worker-2' ]"
+# Number of physical CPUs to use for parallelism
+export THREADS_LIMIT=20
+# Path to kubeconfig
+export KUBECONFIG_PATH=/root/.kube/config
+# Timeout in seconds (0 or negative means no timeout)
+export TIMEOUT=7200
+# Path for logs (YAML files and benchmark_runner.log)
+export LOGS_PATH=/tmp/benchmark-runner-run-artifacts
+```
+
+### Example Script for scaling Fedora VMs
+
+```bash
+#!/bin/bash
+
+# Load Fedora VMs including verification (verify login page)
+sudo podman run --rm -it \
+  -e WORKLOAD='bootstorm_vm' \
+  -e KUBEADMIN_PASSWORD="${KUBEADMIN_PASSWORD}" \
+  -e RUN_ARTIFACTS_URL="${RUN_ARTIFACTS_URL}" \
+  -e SAVE_ARTIFACTS_LOCAL=True \
+  -e SCALE="${BOOTSTORM_SCALE}" \
+  -e FEDORA_CONTAINER_DISK="${FEDORA_CONTAINER_DISK}" \
+  -e SCALE_NODES="${SCALE_NODES}" \
+  -e THREADS_LIMIT="${THREADS_LIMIT}" \
+  -e RUN_STRATEGY=True \
+  -e TIMEOUT="${TIMEOUT}" \
+  -e log_level="INFO" \
+  -v "${LOGS_PATH}:${LOGS_PATH}" \
+  -v "${KUBECONFIG_PATH}:${KUBECONFIG_PATH}" \
+  quay.io/ebattat/benchmark-runner:latest
+```
+Verify the running VMs and get bootstorm_time(millisecond):
+```
+$ oc get vmi -n benchmark-runner
+$ grep bootstorm_time benchmark_runner.log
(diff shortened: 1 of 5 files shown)


docs/source/scaleVM.md
```diff
@@ -0,0 +1,165 @@
...
+- Scale Windows 11 VMs using **ODF-backed storage** (via DataVolume)
+- Parallel VM loading across specified worker nodes
+- Optional metrics collection via **Elasticsearch**
+- Logs and artifacts are stored locally for post-run analysis
+
+---
+
+## 🧪 Prerequisites
+
+- OpenShift cluster with access to worker nodes
```
RobertKrawitz, line 19: What kind of access?  It shouldn't rely on direct ssh access to client nodes, if that's what you mean.

I think that what you really mean is admin access to the cluster.

docs/source/scaleVM.md
```diff
@@ -0,0 +1,165 @@
...
+- [`podman`](https://podman.io) installed
+- Access to the [benchmark-runner](https://quay.io/repository/ebattat/benchmark-runner) container image
+- Properly configured `KUBECONFIG` for cluster access
```
RobertKrawitz, line 22: That's implicit in "OpenShift cluster with access to worker nodes" unless there's something more specific that's needed, in which case it needs to be spelled out.

docs/source/scaleVM.md
```diff
@@ -0,0 +1,165 @@
...
+
+---
+
+## ⚙️ Parameters
+
+### Common Parameters
+| Variable | Description |
+|----------|-------------|
+| `KUBEADMIN_PASSWORD` | OpenShift admin password |
```
RobertKrawitz, line 31: If needed.

docs/source/scaleVM.md
```diff
@@ -0,0 +1,165 @@
...
+| `KUBECONFIG_PATH` | Path to your kubeconfig (e.g. `/root/.kube/config`) |
```
RobertKrawitz, line 32: Kubernetes (including OpenShift) treats `$HOME/.kube/config` as a default, so it's OK if KUBECONFIG is not set.  And the parameter should be named KUBECONFIG.

docs/source/scaleVM.md
```diff
@@ -0,0 +1,165 @@
...
+| `SCALE_NODES` | List of target worker nodes (e.g. `"[ 'worker-0', 'worker-1' ]"`) |
+| `THREADS_LIMIT` | Number of physical CPUs to use for parallelism |
+| `TIMEOUT` | Maximum duration (in seconds) for the run |
```
RobertKrawitz, line 35: Presumably 0 or negative means no timeout, in which case that should be specified here.

docs/source/scaleVM.md
```diff
@@ -0,0 +1,165 @@
...
+| `RUN_ARTIFACTS_URL` | (Optional) Remote URL for saving logs |
+| `SAVE_ARTIFACTS_LOCAL` | If true, saves logs locally to `/tmp/benchmark-runner-run-artifacts` |
```
RobertKrawitz, line 37: It wouldn't be a bad idea for this to optionally be a path to save the logs to.  Although looking further down it's possible to change this by means of volume mapping with podman.

docs/source/scaleVM.md
```diff
@@ -0,0 +1,165 @@
...
+
+---
+
+## 🧰 Fedora VM BootStorm (Ephemeral Storage)
+
+This mode loads Fedora VMs using ephemeral storage and verifies their readiness via the login page.
```
RobertKrawitz, line 43: It should also be possible to do this with ODF.

It would be desirable to be able to use other kinds of volumes (such as localstorage), but it's fine for that to be for the future.

docs/source/scaleVM.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Fedora-Specific Parameters
+
+| Variable | Description                                                |
+|----------|------------------------------------------------------------|
+| `BOOTSTORM_SCALE` | Number of Fedora VMs per worker node                       |
+| `FEDORA_CONTAINER_DISK` | (Optional) Fedora .qcow2 image from quay.io (default: Fedora 37) |
```
RobertKrawitz, line 50: The default should be the actual locator of the image, not the description.

docs/source/scaleVM.md
```diff
@@ -0,0 +1,165 @@
...
+  -e TIMEOUT="${TIMEOUT}" \
+  -e log_level="INFO" \
+  -v "${KUBECONFIG_PATH}:${KUBECONFIG_PATH}" \
+  -v /tmp/benchmark-runner-run-artifacts:/tmp/benchmark-runner-run-artifacts \
+  --privileged quay.io/ebattat/benchmark-runner:latest
+```
+
+## 🧰 Windows VM Bootstorm (ODF Storage)
+
+This mode loads Windows VMs using ODF persistent volumes and validates successful login.
```
RobertKrawitz, line 76: It should (probably) also be possible to do this with ephemeral storage.

docs/source/scaleVM.md
```diff
@@ -0,0 +1,165 @@
...
+  -e SAVE_ARTIFACTS_LOCAL=True \
+  -e SCALE="${WINDOWS_SCALE}" \
+  -e SCALE_NODES="${SCALE_NODES}" \
+  -e THREADS_LIMIT="${THREADS_LIMIT}" \
+  -e WINDOWS_URL="${WINDOWS_URL}" \
+  -e RUN_STRATEGY=True \
+  -e TIMEOUT="${TIMEOUT}" \
+  -e log_level="INFO" \
+  -v "${KUBECONFIG_PATH}:${KUBECONFIG_PATH}" \
+  -v /tmp/benchmark-runner-run-artifacts:/tmp/benchmark-runner-run-artifacts \
```
RobertKrawitz, line 103: This is how it could be saved somewhere else locally.

docs/source/scaleVM.md
```diff
@@ -0,0 +1,165 @@
...
+  --privileged quay.io/ebattat/benchmark-runner:latest
```
RobertKrawitz, line 104: Why does this need to be privileged?  That requires someone else to trust an image not under their control.

RobertKrawitz (changes requested):

docs/source/scaleVM.md
ebattat: Remove it

docs/source/scaleVM.md
ebattat: Its is required, without it it will not work

docs/source/scaleVM.md
ebattat: Update it to $HOME/.kube/config

docs/source/scaleVM.md
ebattat: Added it

docs/source/scaleVM.md
ebattat: Currently its support only ephemeral storage, we can add ODF support in the future.

docs/source/scaleVM.md
ebattat: Fixed

docs/source/scaleVM.md
ebattat: Currently its support only ODF storage, we can add ephemeral support in the future.

docs/source/scaleVM.md
ebattat: I can try to run it w/o it.

docs/source/scaleVM.md
RobertKrawitz: If the KUBECONFIG is already logged in as kubeadmin, this should not be needed.

docs/source/scaleVM.md
RobertKrawitz: OK, we should have an issue for it (restricting Windows to one storage type and Fedora to another seems arbitrary).

docs/source/scaleVM.md
RobertKrawitz: Yes, see above.

docs/source/scaleVM.md
RobertKrawitz: Please do, that would be greatly preferable.

docs/source/scaleVM.md
```diff
@@ -0,0 +1,163 @@
...
+  -e THREADS_LIMIT="${THREADS_LIMIT}" \
+  -e WINDOWS_URL="${WINDOWS_URL}" \
+  -e RUN_STRATEGY=True \
+  -e TIMEOUT="${TIMEOUT}" \
+  -e log_level="INFO" \
+  -e ELASTICSEARCH="${ELASTICSEARCH}" \
+  -e ELASTICSEARCH_PORT="${ELASTICSEARCH_PORT}" \
+  -v "${KUBECONFIG_PATH}:${KUBECONFIG_PATH}" \
+  -v /tmp/benchmark-runner-run-artifacts:/tmp/benchmark-runner-run-artifacts \
+  --privileged quay.io/ebattat/benchmark-runner:latest
```
RobertKrawitz, line 162: Same, see if it can be done w/o privilege.

ebattat: removed privilege

ebattat:
@RobertKrawitz, pls let me know if there is more comments.

RobertKrawitz (approved):

tests/unittest/benchmark_runner/common/oc/test_oc.py
```diff
@@ -7,58 +7,66 @@
...
+        with pytest.raises(YAMLNotExist):
+            oc.create_async(yaml='stressng1.yaml')
 
 
-def dummy_long_uuid(cls, *args, **kwargs):
-    # Your custom testing override
-    return 'bb2be20e-e4e8-57ed-a7c0-efde559b1ce2'
+# Dummy UUID for testing
+def dummy_long_uuid(self, workload):
+    return 'bb2be20e-92d5-4f10-a6ae-94f71f479d12'
```
RobertKrawitz, line 20: Out of curiosity, why did you change this?

tests/unittest/benchmark_runner/common/oc/test_oc.py
```diff
@@ -7,58 +7,66 @@
...
-        oc = OC()
-        assert oc._OC__get_short_uuid(workload='stressng_pod') == 'bb2be20e'
+        oc = OC(kubeadmin_password="dummy-password")  # Avoid LoginFailed
+        short_uuid = oc._OC__get_short_uuid(workload='stressng_pod')
+        assert short_uuid == 'bb2be20e'
+
 
+@patch('benchmark_runner.common.oc.oc.SingletonOCLogin')  # patch login
+def test_check_all_nodes_status_ready(mock_login):
+    oc = OC(kubeadmin_password="dummy-password")  # prevent LoginFailed
```
RobertKrawitz, line 36: A unit test shouldn't need a password at all, since it isn't actually logging in.

RobertKrawitz (changes requested):

tests/unittest/benchmark_runner/common/oc/test_oc.py
ebattat: Thats the reason I have added a patch for benchmark_runner.common.oc.oc.SingletonOCLogin

tests/unittest/benchmark_runner/common/oc/test_oc.py
RobertKrawitz: Why would a unit test failure require you to change a GUID?

ebattat: It is just a mock for [get_long_uuid](https://github.com/redhat-performance/benchmark-runner/blob/129df6a59120ec5e4e32b33667def8445e97c585/tests/unittest/benchmark_runner/common/oc/test_oc.py#L28)

RobertKrawitz: I understand, I just wonder why you changed it here as opposed to using the previous one.

docs/source/scaleVM.md
```diff
@@ -0,0 +1,210 @@
...
+
+This mode loads Windows VMs using ODF persistent volumes and validates successful login.
+
+### Windows-Specific Parameters
+
+| Variable | Description                                                                                 |
+|----------|---------------------------------------------------------------------------------------------|
+| `WINDOWS_SCALE` | Number of Windows VMs per worker node, e.g. 10                                              |
+| `WINDOWS_URL` | URL to the Windows .qcow2 image shared via NGINX (e.g., `http://<host>:8083/windows.qcow2`) |
+
```
jeniferh, line 116: Should you note there is essentially no default (we can't provide access to prebuilt Windows qcow2s)

jeniferh (approved):
I don't understand the SingletonOCLogin test change related to this, but overall the VM scale doc is helpful for users including the local log option.

tests/unittest/benchmark_runner/common/oc/test_oc.py
ebattat: I have mocked `dummy_long_uuid` because it used by `dummy_short_uuid`

docs/source/scaleVM.md
ebattat: I have added more clear description:
`| `WINDOWS_URL` | *(No default, must be provided by the user)* URL to the Windows `.qcow2` image shared via NGINX (e.g., `http://<host>:8083/windows.qcow2`) |`

tests/unittest/benchmark_runner/common/oc/test_oc.py
ebattat: Revert it back to ` bb2be20e-e4e8-57ed-a7c0-efde559b1ce2`

RobertKrawitz (approved):

PR #1019, opened 2025-04-29
Fix upgrade time parsing

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Convert the string to a float first, then to an int 
e.g.: `'5272.361' → float('5272.361') → 5272.361`

Upgrade times come from groovy code and are passed in to the container as environment variables.

### DIFF
--- a/benchmark_runner/workloads/workloads_operations.py
+++ b/benchmark_runner/workloads/workloads_operations.py
@@ -468,7 +468,7 @@ class WorkloadsOperations:
         if ocp_resource_install_minutes_time != 0:
             bm_operations = BareMetalOperations(user=self._environment_variables_dict.get('provision_user', ''))
             ocp_install_minutes_time = bm_operations.get_ocp_install_time()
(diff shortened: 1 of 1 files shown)


benchmark_runner/workloads/workloads_operations.py
```diff
@@ -468,7 +468,7 @@ def update_ci_status(self, status: str, ci_minutes_time: int, benchmark_runner_i
...
             ocp_install_minutes_time = bm_operations.get_ocp_install_time()
-        metadata.update({'status': status, 'status#': status_dict[status], 'ci_minutes_time': ci_minutes_time, 'benchmark_runner_id': benchmark_runner_id, 'benchmark_operator_id': benchmark_operator_id, 'benchmark_wrapper_id': benchmark_wrapper_id, 'ocp_install_minutes_time': ocp_install_minutes_time, 'ocp_resource_install_minutes_time': ocp_resource_install_minutes_time, 'upgrade_masters_duration_seconds': int(self._upgrade_masters_duration_seconds), 'upgrade_workers_duration_seconds': int(self._upgrade_workers_duration_seconds)})
+        metadata.update({'status': status, 'status#': status_dict[status], 'ci_minutes_time': ci_minutes_time, 'benchmark_runner_id': benchmark_runner_id, 'benchmark_operator_id': benchmark_operator_id, 'benchmark_wrapper_id': benchmark_wrapper_id, 'ocp_install_minutes_time': ocp_install_minutes_time, 'ocp_resource_install_minutes_time': ocp_resource_install_minutes_time, 'upgrade_masters_duration_seconds': int(float(self._upgrade_masters_duration_seconds)), 'upgrade_workers_duration_seconds': int(float(self._upgrade_workers_duration_seconds))})
```
RobertKrawitz, line 471: This looks fine; my only question is how this would get set to such a value in the first place.  It appears that this comes from the environment, rather than being set from a measured interval; why would it be set to something like this?

ebattat: I got it from Jenkins groovy code as a string and time in seconds

RobertKrawitz (approved):

PR #1023, opened 2025-05-07
Upload krknhub logs into google drive

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Upload krknhub logs into google drive or S3

### DIFF
--- a/benchmark_runner/krkn_hub/krknhub_workloads.py
+++ b/benchmark_runner/krkn_hub/krknhub_workloads.py
@@ -44,6 +44,12 @@ class KrknHubWorkloads(WorkloadsOperations):
         """
         if self._enable_prometheus_snapshot:
             self.end_prometheus()
+        if self._google_drive_shared_drive_id:
+            self.upload_run_artifacts_to_google_drive()
+        elif self._endpoint_url:
+            self.upload_run_artifacts_to_s3()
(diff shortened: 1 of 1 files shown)


benchmark_runner/krkn_hub/krknhub_workloads.py
```diff
@@ -44,6 +44,12 @@ def finalize_workload(self):
         """
         if self._enable_prometheus_snapshot:
             self.end_prometheus()
+        if self._google_drive_shared_drive_id:
+            self.upload_run_artifacts_to_google_drive()
+        elif self._endpoint_url and not self._google_drive_shared_drive_id:
```
RobertKrawitz, line 49: Don't need the "and not self.__google_drive_shared_drive_id" because that's already covered by the previous case.  Otherwise everything is good.

RobertKrawitz (changes requested):

benchmark_runner/krkn_hub/krknhub_workloads.py
ebattat: fixed

RobertKrawitz (approved):

PR #1025, opened 2025-05-07
Upload into elastic verification each vm  verification and failure

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Currently, there is one row per VM verification in Elasticsearch per test cycle.
This fix uploads a verification indicator for each VM, as well as a single row listing the failed VMs in case of a failure.

### DIFF
--- a/benchmark_runner/workloads/bootstorm_vm.py
+++ b/benchmark_runner/workloads/bootstorm_vm.py
@@ -123,15 +123,15 @@ class BootstormVM(WorkloadsOperations):
                 self._data_dict.update(self._prometheus_result)
             total_run_time = self._get_bootstorm_vm_total_run_time()
             self._data_dict.update({'total_run_time': total_run_time})
-            # Google drive run_artifacts_url folder path
-            if self._google_drive_path and self.get_run_artifacts_google_drive():
-                self._data_dict.update({'run_artifacts_url': self.get_run_artifacts_google_drive()})
-            if self._es_host:
-                # upload several run results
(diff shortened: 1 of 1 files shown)


benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -292,19 +292,19 @@ def _verify_virtctl_vms(self, delay=10):
...
-                if self._es_host:
-                    self._data_dict.update({'run_artifacts_url': self.get_run_artifacts_google_drive()})
-                    # upload several run results
-                    self._upload_to_elasticsearch(index=self._es_index, kind=self._kind, status=self._status,result=self._data_dict)
-                    # verify that data upload to elastic search according to unique uuid
-                    self._verify_elasticsearch_data_uploaded(index=self._es_index, uuid=self._uuid)
+                    # Upload artifacts
+                    if self._google_drive_shared_drive_id:
+                        self.upload_run_artifacts_to_google_drive()
+                    elif self._endpoint_url and not self._google_drive_shared_drive_id:
```
RobertKrawitz, line 298: Don't need the "and not self._google_drive_shared_drive_id" because that's already covered above.

RobertKrawitz (changes requested):
One small change.

benchmark_runner/workloads/bootstorm_vm.py
ebattat: fixed

RobertKrawitz (approved):

PR #1026, opened 2025-05-12
Upload to google drive only in case of error

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Currently, we upload to Google Drive on every run. 
We should upload to Google Drive only in case of an error and avoid cluttering it with logs from every run.

### DIFF
--- a/benchmark_runner/workloads/bootstorm_vm.py
+++ b/benchmark_runner/workloads/bootstorm_vm.py
@@ -123,9 +123,6 @@ class BootstormVM(WorkloadsOperations):
                 self._data_dict.update(self._prometheus_result)
             total_run_time = self._get_bootstorm_vm_total_run_time()
             self._data_dict.update({'total_run_time': total_run_time})
-        # Google drive run_artifacts_url folder path
-        if self._google_drive_path and self.get_run_artifacts_google_drive():
-            self._data_dict.update({'run_artifacts_url': self.get_run_artifacts_google_drive()})
         if self._es_host:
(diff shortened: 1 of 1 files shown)


ebattat:
I will create a new PR including this small fix

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -123,9 +123,6 @@ def _finalize_vm(self):
                 self._data_dict.update(self._prometheus_result)
             total_run_time = self._get_bootstorm_vm_total_run_time()
             self._data_dict.update({'total_run_time': total_run_time})
-        # Google drive run_artifacts_url folder path
-        if self._google_drive_path and self.get_run_artifacts_google_drive():
-            self._data_dict.update({'run_artifacts_url': self.get_run_artifacts_google_drive()})
```
RobertKrawitz, line 128: Where is this done in the error case?

RobertKrawitz (commented):
You've closed and reopened it.  I don't have any major objection to this but it looks like you've had second thoughts about it.  Please make sure that this is really what you want to do.

ebattat:
Its cause to duplicate folder in google drive.
I plan to change it in the next PR so that we create only one folder during initialization and then upload into it at the end, just like we do for S3.

ebattat:
Closing, I will fix in other PR

PR #1027, opened 2025-05-12
Using Virtctl instead of ssh

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Using Virtctl instead of SSH is faster, simpler, and eliminates the need for a complex flow involving exposing the VM using an IP and port.
** keep the ssh/ vm exposed methods

### DIFF
--- a/benchmark_runner/common/oc/oc.py
+++ b/benchmark_runner/common/oc/oc.py
@@ -1289,9 +1289,27 @@ class OC(SSH):
         else:
             return self.run(ssh_vm_cmd)
 
-    def get_virtctl_vm_status(self, vm_name: str = '', namespace: str = environment_variables.environment_variables_dict['namespace']):
+    def wait_for_vm_access(self, vm_name: str = '', namespace: str = environment_variables.environment_variables_dict['namespace'],
+                        timeout: int = SHORT_TIMEOUT):
         """
-        This method returns True when the VM is active and an error message when it is not, using virtctl protocol
+        This method waits for VM to be accessible via virtctl ssh
+        :param vm_name:
+        :param timeout:
+        :return:
+        """
+        current_wait_time = 0
+        while timeout <= 0 or current_wait_time <= timeout:
+            check_virtctl_vm_cmd = f"virtctl ssh --local-ssh=true --local-ssh-opts='-o BatchMode=yes' --local-ssh-opts='-o PasswordAuthentication=no' --local-ssh-opts='-o ConnectTimeout=2' root@{vm_name} -n {namespace} 2>&1 |egrep 'denied|verification failed'  && echo 'True' || echo 'False'"
+            if 'True' in self.run(check_virtctl_vm_cmd):
+                return 'True'
+            # sleep for x seconds
+            time.sleep(OC.SLEEP_TIME)
+            current_wait_time += OC.SLEEP_TIME
+        raise VMStateTimeout(vm_name=vm_name, state='virtctl')
+
+    def get_vm_access(self, vm_name: str = '', namespace: str = environment_variables.environment_variables_dict['namespace']):
+        """
+        This method returns True when the VM is access and an error message when it is not, using virtctl protocol
         :param vm_name:
         :param namespace:
         :return: virtctl_status 'True' if successful, or an error message if it fails.
@@ -1312,7 +1330,7 @@ class OC(SSH):
         :return:
         """
         namespace = f'-n {namespace}' if namespace else ''
-        command = f"{self._cli} get vmi {vm_name} {namespace} -o jsonpath={{.metadata.labels.'kubevirt\\.io/nodeName'}}"
+        command = f"{self._cli} get vmi {vm_name} {namespace} -o jsonpath=\"{{.status.nodeName}}\""
 
         try:
             result = self.run(command)
--- a/benchmark_runner/workloads/bootstorm_vm.py
+++ b/benchmark_runner/workloads/bootstorm_vm.py
@@ -67,6 +67,19 @@ class BootstormVM(WorkloadsOperations):
         self._bootstorm_start_time[vm_name] = time.time()
 
     @logger_time_stamp
+    def _wait_vm_access(self, vm_name: str):
+        """
+        This method verifies virtctl into VM and return vm node in success or False if failed
+        @return:
+        """
+        if self._oc.get_vm_node(vm_name=vm_name):
+            vm_node = self._oc.get_vm_node(vm_name=vm_name)
+            if self._oc.wait_for_vm_access(vm_name=vm_name):
+                logger.info(f"Successfully virtctl into VM: '{vm_name}' in node {vm_node} ")
+            return vm_node
+        return False
+
+    @logger_time_stamp
     def _wait_ssh_vm(self, vm_name: str):
         """
         This method verifies ssh into VM and return vm node in success or False if failed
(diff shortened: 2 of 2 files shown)


benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -92,7 +106,7 @@ def _get_bootstorm_vm_elapsed_time(self, vm_name: str, vm_node: str) -> dict:
         """
         if vm_node:
             delta = round((time.time() - self._bootstorm_start_time[vm_name]) * self.MILLISECONDS, 3)
-            data = {'vm_name': vm_name, 'node': vm_node, 'bootstorm_time': delta, 'vm_ssh': int(bool(vm_node)),}
+            data = {'vm_name': vm_name, 'node': vm_node, 'bootstorm_time': delta, 'virtctl_vm': int(bool(vm_node)),}
```
RobertKrawitz, line 109: Rather than name this "vm_ssh" or "virtctl_vm", name it something like "vm_status" to reflect its function.

benchmark_runner/common/oc/oc.py
```diff
@@ -1289,6 +1289,24 @@ def get_vm_ssh_status(self, node_ip: str = '', vm_node_port: str = ''):
         else:
             return self.run(ssh_vm_cmd)
 
+    def wait_for_vm_virtctl(self, vm_name: str = '', namespace: str = environment_variables.environment_variables_dict['namespace'],
```
RobertKrawitz, line 1292: Create a function `get_vm_status` that hides the method used to actually do it.  You shouldn't need to change any other code to switch between `ssh` and `virtctl ssh`.

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -66,6 +66,20 @@ def _set_bootstorm_vm_start_time(self, vm_name: str = ''):
...
+        @return:
+        """
+        # wait till vm virtctl login
+        if self._oc.get_vm_node(vm_name=vm_name):
+            vm_node = self._oc.get_vm_node(vm_name=vm_name)
+            if self._oc.wait_for_vm_virtctl(vm_name=vm_name):
+                logger.info(f"Successfully virtctl into VM: '{vm_name}' in node {vm_node} ")
+            return vm_node
+        return False
+
```
RobertKrawitz, line 82: Again, name it something generic so that the underlying mechanism isn't exposed and you can change it in the future without having to change anything outside of `oc.py`.

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -143,7 +157,7 @@ def _run_vm(self):
         self._set_bootstorm_vm_first_run_time()
         self._set_bootstorm_vm_start_time(vm_name=self._vm_name)
         self._virtctl.start_vm_sync(vm_name=self._vm_name)
-        self.vm_node = self._wait_ssh_vm(vm_name=self._vm_name)
+        self.vm_node = self._wait_virtctl_vm(vm_name=self._vm_name)
```
RobertKrawitz, line 160: Change to `_wait_for_vm_status` or similar without specifying the mechanism in the name.

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -322,7 +336,7 @@ def _run_vm_scale(self, vm_num: str):
             if not self._run_strategy:
                 self._virtctl.start_vm_async(vm_name=f'{self._workload_name}-{self._trunc_uuid}-{vm_num}')
             self._virtctl.wait_for_vm_status(vm_name=vm_name, status=VMStatus.Running)
-            vm_node = self._wait_ssh_vm(vm_name)
+            vm_node = self._wait_virtctl_vm(vm_name)
```
RobertKrawitz, line 339: See previous.

RobertKrawitz (changes requested):

benchmark_runner/common/oc/oc.py
ebattat: Its 2 different approaches, I dont want to have any relation between it.

benchmark_runner/workloads/bootstorm_vm.py
ebattat: I dont think we should keep the 2 approaches, I plan to shift all to virtctl.

benchmark_runner/workloads/bootstorm_vm.py
ebattat: I prefer to use the right term virtctl instead of status.
If there is any issue, it will be more informative.

ebattat:
@RobertKrawitz, 
I could change it to status, but I think that's too general. 
If there's an issue with it later, we'll need to check how we actually check the VM status.

benchmark_runner/workloads/bootstorm_vm.py
RobertKrawitz: OK.

benchmark_runner/common/oc/oc.py
RobertKrawitz: Its two different approaches, but the end result is the same and nothing outside of this file should have any reason to know how we retrieve the status.

benchmark_runner/workloads/bootstorm_vm.py
RobertKrawitz: We don't need to keep both approaches.  We should simply use one name that's visible outside of the implementation.  You should not have to change any code outside of this file.

If there are more places that need to be changed, please make all of the necessary changes in one PR.

benchmark_runner/workloads/bootstorm_vm.py
RobertKrawitz: OK.

RobertKrawitz:
> @RobertKrawitz, I could change it to status, but I think that's too general. If there's an issue with it later, we'll need to check how we actually check the VM status.

You might name it "access" or similar, as whether you use ssh or virtctl is immaterial.  You want to check that you are able to access the VM; the mechanism does not matter.

benchmark_runner/workloads/bootstorm_vm.py
ebattat: I see that we already use wait_for_vm_status if it stop/ready/run.
Do u think we should use the same term for virtctl ?

benchmark_runner/common/oc/oc.py
```diff
@@ -1289,9 +1289,27 @@ def get_vm_ssh_status(self, node_ip: str = '', vm_node_port: str = ''):
...
+                return 'True'
+            # sleep for x seconds
+            time.sleep(OC.SLEEP_TIME)
+            current_wait_time += OC.SLEEP_TIME
+        raise VMStateTimeout(vm_name=vm_name, state='virtctl')
+
+    def get_vm_access(self, vm_name: str = '', namespace: str = environment_variables.environment_variables_dict['namespace']):
         """
-        This method returns True when the VM is active and an error message when it is not, using virtctl protocol
+        This method returns True when the VM is access and an error message when it is not, using virtctl protocol
```
RobertKrawitz, line 1312: "accessible"

RobertKrawitz (approved):
One trivial typo in a comment is not enough to withhold approval.

PR #1028, opened 2025-05-13
Uploaded logs to Google Drive once

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Currently we upload to google drive every several time when there are several vms.
This PR fix is verified that there is 1 google drive folder per day run and upload the logs into it.
In addition, I have split create_drive_folder_url and get_drive_folder_url to 2 different methods.

### DIFF
--- a/benchmark_runner/common/google_drive/google_drive_operations.py
+++ b/benchmark_runner/common/google_drive/google_drive_operations.py
@@ -9,7 +9,7 @@ from googleapiclient.http import MediaFileUpload
 from google.auth.transport.requests import Request
 from google_auth_oauthlib.flow import InstalledAppFlow
 from benchmark_runner.common.logger.logger_time_stamp import logger  # Added logger import
-
+from google_drive_exceptions import FolderNotCreated
 # Define the scope
 SCOPES = ['https://www.googleapis.com/auth/drive']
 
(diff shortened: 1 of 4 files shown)


benchmark_runner/workloads/workloads_operations.py
```diff
@@ -524,6 +533,8 @@ def initialize_workload(self):
         self._template.generate_yamls(scale=str(self._scale), scale_nodes=self._scale_node_list, redis=self._redis, thread_limit=self._threads_limit)
         if self._enable_prometheus_snapshot:
             self.start_prometheus()
+        if self._google_drive_path:
+            self.find_or_create_run_artifacts_google_drive()
```
RobertKrawitz, line 537: find_or_create_run_artifacts_drive returns the new drive, but this doesn't use the returned value.

benchmark_runner/common/google_drive/google_drive_operations.py
```diff
@@ -241,16 +241,38 @@ def get_drive_folder_url(self, folder_path, parent_folder_id):
...
+        """
+        Find an existing folder by path or create it. Retries in case of failure.
+        :param folder_path: The full folder path to retrieve or create.
+        :param parent_folder_id: The starting parent folder ID.
+        :param retries: Number of retry attempts on failure.
+        :param delay: Initial delay between retries in seconds.
+        :return: The Google Drive URL for the folder, or None on failure.
+        """
+        for attempt in range(1, retries + 1):
+            folder_id = self.get_folder_id_by_path(folder_path, parent_folder_id)
```
RobertKrawitz, line 255: The race condition still exists; the check for it could fail, but something else could create it before we get around to it.  It's likely a rather narrow race condition, but I'd be very surprised if Google Drive does not offer a way to create a folder only if it does not exist.  Please try to look for that way.

ebattat: I tested it and it's working properly. I call it once before all the runs, so it's created only once, and that solved the issue.

benchmark_runner/workloads/workloads_operations.py
ebattat: I revert it back to create

benchmark_runner/common/google_drive/google_drive_operations.py
RobertKrawitz: It will work properly if you can guarantee that only one instance is running at a time.  The problem is if multiple instances are running and both want to create the same folder name.  As I said, a small window, but it's not a great idea to assume that we're the only use case.

benchmark_runner/workloads/workloads_operations.py
```diff
@@ -373,6 +373,15 @@ def get_run_artifacts_google_drive(self):
...
         return self._google_drive_operation.get_drive_folder_url(folder_path=run_artifacts_hierarchy, parent_folder_id=self._google_drive_shared_drive_id)
 
+    def create_run_artifacts_google_drive(self):
+        """
+        This method creates google drive run artifacts folder path
+        :return:
+        """
+        workload = self._workload.replace('_', '-')
+        run_artifacts_hierarchy = self._get_run_artifacts_hierarchy(workload_name=workload)
+        return self._google_drive_operation.create_drive_folder_url(folder_path=run_artifacts_hierarchy, parent_folder_id=self._google_drive_shared_drive_id)
```
RobertKrawitz, line 383: This returns a value, but the caller below (at line 537) does not use the return value as it should.

RobertKrawitz (approved):

PR #1035, opened 2025-05-21
Passed dynamic arguments into the Dockerfile

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

1. Add new passing argument VIRTCTL_VERSION  into Dockerfile  
2. Add missing argument OCP_CLIENT_VERSION into Build_runner.yaml
3. Update Dockerfile with the new argument VIRTCTL_VERSION

PR #1036, opened 2025-05-21
Update benchmark-runner version manually due to build failure

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Update the benchmark-runner version due to a failed [manual run](https://github.com/redhat-performance/benchmark-runner/actions/runs/15136753760) due to missing client version, fixed in [PR](https://github.com/redhat-performance/benchmark-runner/pull/1035)

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,7 +1,7 @@
 [bumpversion]
 commit = False
 tag = True
-current_version = 1.0.812
+current_version = 1.0.813
 tag_name = v{current_version}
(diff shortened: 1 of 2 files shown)


RobertKrawitz (approved):
I'm approving this because it's necessary to unstick the pipeline, but even in the event of a build failure it should not be necessary to do this.  Please open an issue to fix the requirement to manually update the version in the event of a build failure.

PR #1040, opened 2025-05-28
Check that VMs are running before attempting to verify access

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Check if there are running VMs before entering to the verification method _verify_vms_access

### DIFF
--- a/benchmark_runner/workloads/bootstorm_vm.py
+++ b/benchmark_runner/workloads/bootstorm_vm.py
@@ -236,12 +236,16 @@ class BootstormVM(WorkloadsOperations):
         self._oc.save_describe_yml(vm_name, str(vm_access).lower() == 'true', output_dir=self._run_artifacts_path)
         return_dict[vm_name] = vm_access
 
-    def _verify_vms_access_in_parallel(self, vm_names):
+    def _verify_vms_access_in_parallel(self, vm_names: list):
         """
-        This method verifies VM access in parallel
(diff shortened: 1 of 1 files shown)


RobertKrawitz (changes requested):

ebattat:
I have added another check for vm_names in _verify_vms_access_in_parallel method

benchmark_runner/workloads/bootstorm_vm.py
```diff
@@ -236,87 +236,86 @@ def _verify_single_vm_access_wrapper(self, vm_name, return_dict):
...
-        :return:
+        Verifies access to a list of VMs in parallel using subprocesses.
+
+        :param vm_names: List of VM names to verify.
+        :return: List of VM names that failed the access verification.
         """
-        failure_vms = []
-        manager = Manager()
-        return_dict = manager.dict()
+        if vm_names:
```
RobertKrawitz, line 246: Need to return None if there are no VM names.

RobertKrawitz (changes requested):

benchmark_runner/workloads/bootstorm_vm.py
ebattat: Done,
```
if not vm_names:
    return []
```

PR #1042, opened 2025-06-03
Fix '-ec' '-rc' version Grafana display

ebattat:
## Type of change
Note: Fill **x** in []
- [x] bug
- [ ] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Currently, 'ec' and 'rc' versions are displayed the same in Grafana because they must be represented as numbers without any characters.
For example: 41904 could mean either 4.19.0-ec.4 or 4.19.0-rc.4.
To resolve this conflict, the solution is to represent 'e' as '0' and 'r' as '1'.
In addition, I fixed the Graffonet code by .replace("e", "0") .replace("r", "1")  and removed duplicate code.


benchmark_runner/common/grafana/update_grafana_latest_value_mappings.py
```diff
@@ -89,8 +89,11 @@ def update_value_mappings_last_versions(self, last_versions: dict):
 
         num = 1
         for key, value in last_versions.items():
-            if not self.value_mappings.get(key):
-                self.value_mappings[key] = {"index": int(max_index) + num, "text": value}
+            if not self.value_mappings.get(key) or self.value_mappings[key]['text'] != value:
+                # Represent 'rc' versions using '1' to avoid conflict with 'ec'
+                if 'r' in value:
+                    value = ''.join('1' if c == 'r' else c for c in value if c not in '.cfe-')
+                self.value_mappings[value] = {"index": int(max_index) + num, "text": value}
```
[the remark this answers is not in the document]
ebattat: We plan to run only ec/rc/ga versions, no plan to run nightly or fc.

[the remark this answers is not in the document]
ebattat: I have updated it accordingly:

```
if '-ec' in value:
  ...
# Represent 'rc' versions using extra '1' (Grafana panel supports only numbers)
elif '-rc' in value:
 ...
# Represent 'ga' (Default)
else:
 ...
```

benchmark_runner/common/grafana/update_grafana_latest_value_mappings.py
```diff
@@ -45,7 +45,16 @@ def get_last_elasticsearch_versions(self, last_es_fetch_days=LAST_ES_FETCH_DAYS)
...
+                    if '-ec.' in value:
+                        full_version = value
+                        value = ''.join('0' if c == 'e' else c for c in value if c not in '.cfr-')
+                        new_versions[value] = full_version
+                    elif '-rc.' in value:
+                        full_version = value
+                        value = ''.join('1' if c == 'r' else c for c in value if c not in '.cfe-')
+                        new_versions[value] = full_version
+                    else:
+                        new_versions[value.translate(str.maketrans('', '', '.-rcef'))] = value
```
RobertKrawitz, line 57: This code is duplicated in two places.  Please create a single function that encapsulates both, and simplify the logic.  You don't need that for loop here at all.

ebattat: I have have merged it to 1 method.

benchmark_runner/common/grafana/update_grafana_latest_value_mappings.py
```diff
@@ -26,6 +26,18 @@ def __init__(self, main_libsonnet_path: str):
...
+    def _normalize_version_for_grafana(value: str):
+        """
+         Normalizes a version string for Grafana, convert 'e' to '0', 'r' to '1'
+        """
+        if '-ec.' in value:
+            return ''.join('0' if c == 'e' else c for c in value if c not in '.cfr-')
+        elif '-rc.' in value:
+            return ''.join('1' if c == 'r' else c for c in value if c not in '.cfe-')
+        else:
+            return None
```
RobertKrawitz, line 39: This code is still very complicated.

I believe a version string is of the form (using regular expressions) of

```
(4)\.([0-9]+)\.([0-9]+)((-ec|-rc)\.([0-9]+))?
```

So we have the following groups if the match is successful:

```
1      Major (should always be 4)
2      Minor
3      Point
4      Qualifier
5         Qualifier type (-ec or -rc)
6         Qualifier version
```
I don't know exactly how the number is built (that needs to be documented), but you really want a function that does this parsing and generates a number.  If the version string does not match the regex, that's an error and you either want to raise an exception or use a defined invalid number.

So assuming that the string is MMMmmmpppQqqq where M is major, m is minor, p is point, Q is qualifier (0=ec, 1=rc, 2=none), and q is qualifier version, the code would look something like this.  It's a bit longer, but a lot easier to read.

```
# <major>.<minor>.<point><optional_qualifier>
# qualifier if present is <-ec|-rc>.<qualifier_version>
m = re.match("(4)\.([0-9]+)\.([0-9]+)((-ec|-rc)\.([0-9]+))?
if m:
    major = int(m[1])
    minor = int(m[2])
    point = int(m[3])
    qualifier = m[5]
    qualifier_version = m[6]
    if qualifier == '-ec':
        qualifier = 0
        qualifier_version = int(qualifier_version)
    elif qualifier == '-rc':
        qualifier = 1
        qualifier_version = int(qualifier_version)
    else:
        qualifier = 2
        qualifier_version = 0
    return '%d%03d%03d%1d%03d' % (major, minor, point, qualifier, qualified_version)
else:
    return -1 # or whatever is your invalid version
```

ebattat: @RobertKrawitz, 
We already have data in Elasticsearch, so I want to make minimal changes to keep supporting the existing data.
I have tested it, and everything is working fine.

ebattat: I dont think its a good idea.
The whole idea of this PR is to parse char to int in order to display it properly in Grafana.

RobertKrawitz: It's not at all clear to me how this operates.  If somebody else had to maintain this code, how would they know what the output should be?  If I give you the following strings, what do you expect the output to be in each case?

- 4.4.0
- 4.4.19
- 4.10.0
- 4.10.20
- 4.19.0-rc.0
- 4.19.0-rc.12
- 4.19.0-ec.1
- 4.19.0-ec.17

benchmark_runner/grafana/perf/jsonnet/main.libsonnet
```diff
@@ -286,7 +286,7 @@ g.dashboard.new('PerfCI-Regression-Summary')
           + elasticsearch.withMetrics([
             elasticsearch.metrics.MetricAggregationWithSettings.Max.withField('ci_minutes_time')
             + elasticsearch.metrics.MetricAggregationWithSettings.Max.withId('1')
-            + elasticsearch.metrics.MetricAggregationWithSettings.Max.settings.withScript('Integer.parseInt(\"0\"+doc["ocp_version.keyword"].value.replace(\".\",\"\").replace(\"r\",\"\").replace(\"c\",\"\").replace(\"f\",\"\").replace(\"-\",\"\").replace(\"e\",\"\").replace(\"c\",\"\").replace(\"r\",\"\"))')
+            + elasticsearch.metrics.MetricAggregationWithSettings.Max.settings.withScript('Integer.parseInt(\"0\"+doc["ocp_version.keyword"].value.replace(\".\",\"\").replace(\"r\",\"1\").replace(\"e\",\"0\").replace(\"c\",\"\").replace(\"f\",\"\").replace(\"-\",\"\"))')
```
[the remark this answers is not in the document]
RobertKrawitz: I understand in general, but you're presumably encoding the -ec, -rc, etc, not just dropping them, correct?

Does the code properly distinguish between 4.11.5 and 4.1.15?

benchmark_runner/common/grafana/update_grafana_latest_value_mappings.py
ebattat: The general purpose is to remove characters from the string because Grafana can only display integers.
To distinguish between RC and EC, I represent them with numbers: r = 1 and e = 0.
Therefore, we don't need to split and rebuild the version, we only need to replace r and c.

benchmark_runner/grafana/perf/jsonnet/main.libsonnet
```diff
@@ -225,7 +225,7 @@ g.dashboard.new('PerfCI-Regression-Summary')
...
... {"index": 417, "text": "4.18.2-11"}, "41904": {"index": 418, "text": "4.19.0-ec.4"}, "419099": {"index": 419, "text": "4.19.0-99"}, "4177": {"index": 420, "text": "4.17.7"}, "4176": {"index": 421, "text": "4.17.6"}, "4188": {"index": 422, "text": "4.18.8"}, "418213": {"index": 423, "text": "4.18.2-13"}, "4190100": {"index": 424, "text": "4.19.0-100"}, "4189": {"index": 425, "text": "4.18.9"}, "418225": {"index": 426, "text": "4.18.2-25"}, "41810": {"index": 427, "text": "4.18.10"}, "41831": {"index": 428, "text": "4.18.3-1"}, "4190117": {"index": 429, "text": "4.19.0-117"}, "41905": {"index": 430, "text": "4.19.0-ec.5"}, "4190124": {"index": 431, "text": "4.19.0-124"}, "41812": {"index": 432, "text": "4.18.12"}, "418312": {"index": 433, "text": "4.18.3-12"}, "41813": {"index": 434, "text": "4.18.13"}, "41847": {"index": 435, "text": "4.18.4-7"}, "4190145": {"index": 436, "text": "4.19.0-145"}, "4190154": {"index": 437, "text": "4.19.0-154"}, "41902": {"index": 438, "text": "4.19.0-rc.2"}, "4190160": {"index": 439, "text": "4.19.0-160"}, "4190170": {"index": 440, "text": "4.19.0-170"}, "4190164": {"index": 441, "text": "4.19.0-164"}, "4190174": {"index": 442, "text": "4.19.0-174"}}
```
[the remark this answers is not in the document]
ebattat: I’ve resolved the conflict, and it’s OK now.
"413004": {"index": 101, "text": "4.13.0-ec.4"},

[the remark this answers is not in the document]
RobertKrawitz: I'm assuming that the purpose of using the specific mapping that you did was because the ID number is an index that you can reverse map to the release in question.  It may not be possible to avoid some incompatible changes, but you should think carefully about what you want the results to be.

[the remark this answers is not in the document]
RobertKrawitz: OK, this is your call.

RobertKrawitz (approved):

PR #1061, opened 2025-07-08
Create VMs only for fast VMs deployment

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Create VMs only for fast VMs deployment.


jeniferh (approved):

ebattat:
@RobertKrawitz, currently, we are deploying VMs with verification, which is time-consuming.
This PR adds the ability to deploy VMs without verification.
So the veification you can do in separate cycle.

PR #1065, opened 2025-07-16
Fence agents remediation operator

ebattat:
## Type of change
Note: Fill **x** in []
- [ ] bug
- [x] enhancement
- [ ] documentation
- [ ] dependencies

## Description

Adding fence-agents-remediation instead of self-node-remediation operator including verification steps


benchmark_runner/common/ocp_resources/nhc_far/template/05_nhc_far_template.yaml
```diff
@@ -0,0 +1,60 @@
...
+        '--lanplus': ''
+      agent: fence_ipmilan
+
+---
+apiVersion: remediation.medik8s.io/v1alpha1
+kind: NodeHealthCheck
+metadata:
+  name: nhc-remediation-far
+spec:
+  minHealthy: 1
```
jeniferh, line 42: We need to ensure only 1 node is remediated at one time 
so minHealthy should be (total nodes -1)  -- and soon they will had a more flexible maxUnhealthy option...

ebattat: I have updated it to `minHealthy: 2`

PR #1186, opened 2026-03-13
Migrate stressng and uperf workloads from benchmark-operator/snafu to native resource creation

arpsharm:
### What changed

- Remove benchmark-operator and snafu/benchmark-wrapper dependencies
- Pod workloads use native Kubernetes Jobs with inline JSON parsers
- VM workloads use native KubeVirt VirtualMachines with cloud-init
- VM results extracted via virtctl SSH/SCP (SSH keys injected via cloud-init)
- SSH/SCP methods added to Virtctl class: `wait_for_vm_workload_completed`, `_ssh_ready`, `_wait_for_file_created`, `_scp_file`
- Extended existing `oc.py` methods (`get_pod`, `wait_for_pod_create`, `delete_vm_sync`) instead of adding new ones
- Added `get_pod_ip`, `get_pod_node`, `get_vmi_ip`, `get_cluster_name` helpers to `oc.py`
- Replaced all `subprocess.run` calls with OC/Virtctl methods
- Added `@typechecked`, `@logger_time_stamp` annotations, moved initializations to `__init__`
- Combined YAML templates (Namespace + ConfigMap/Secret + Job/VM in single file)
- Fixed `stressng_timeout` variable collision with general `timeout` env var
- ES upload handled by benchmark-runner Python (not inside pods/VMs)
- 95th percentile latency with numpy-equivalent linear interpolation
- All ES fields match existing schema for Grafana compatibility
- Prometheus metrics populated
- Added integration tests for OC methods (error handling, timeout, UUID) and Virtctl SSH/SCP
- Updated README.md and podman.md: removed benchmark-operator references, default namespace changed to `benchmark-runner`

PR #1189, opened 2026-04-09
Remove stale CI status params and add explicit vm_user for stressng/uperf

arpsharm:
## Type of change
- [x] bug
- [x] enhancement

## Description
1. Remove stale `benchmark_operator_id` and `benchmark_wrapper_id` from `update_ci_status()` and CI workflows — these were removed in PR #1186 but the callers still passed them, causing `TypeError` at runtime.

2. Add explicit `username='fedora'` to stressng_vm and uperf_vm workloads so they don't rely on the global `vm_user` default (preparation for hammerdb PR #1190 which changes the default to empty).

3. Fix virtctl integration test to pass `username='fedora'` explicitly.

### DIFF
--- a/benchmark_runner/main/main.py
+++ b/benchmark_runner/main/main.py
@@ -213,13 +213,9 @@ def update_ci_status():
     """
     ci_minutes_time = environment_variables_dict.get('ci_minutes_time', '')
     benchmark_runner_id = environment_variables_dict.get('benchmark_runner_id', '')
-    benchmark_operator_id = environment_variables_dict.get('benchmark_operator_id', '')
-    benchmark_wrapper_id = environment_variables_dict.get('benchmark_wrapper_id', '')
     benchmark_runner = Workloads()
     benchmark_runner.update_ci_status(status=ci_status, ci_minutes_time=int(ci_minutes_time),
-                                      benchmark_runner_id=benchmark_runner_id,
-                                      benchmark_operator_id=benchmark_operator_id,
-                                      benchmark_wrapper_id=benchmark_wrapper_id)
+                                      benchmark_runner_id=benchmark_runner_id)
 
 
 @logger_time_stamp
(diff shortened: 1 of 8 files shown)


benchmark_runner/main/main.py
```diff
@@ -213,13 +213,9 @@ def update_ci_status():
...
     ci_minutes_time = environment_variables_dict.get('ci_minutes_time', '')
     benchmark_runner_id = environment_variables_dict.get('benchmark_runner_id', '')
-    benchmark_operator_id = environment_variables_dict.get('benchmark_operator_id', '')
-    benchmark_wrapper_id = environment_variables_dict.get('benchmark_wrapper_id', '')
     benchmark_runner = Workloads()
     benchmark_runner.update_ci_status(status=ci_status, ci_minutes_time=int(ci_minutes_time),
-                                      benchmark_runner_id=benchmark_runner_id,
-                                      benchmark_operator_id=benchmark_operator_id,
-                                      benchmark_wrapper_id=benchmark_wrapper_id)
+                                      benchmark_runner_id=benchmark_runner_id)
```
ebattat, line 218: Is it the only place we use benchmark_operator_id and benchmark_wrapper_id ?

ebattat (commented):
Is it the only place we use it ?

benchmark_runner/main/main.py
arpsharm: Yes it was the only file being directly affected at runtime. However, there were stale references as well set in the yaml files, usually harmless but I removed that from there as well just to be consistent.

ebattat: can u remove it from jenkinsfiles also ?

arpsharm: I don't see any more references for this in Jenkinsfiles. Am I missing something?

ebattat: No, its ok.

ebattat (approved):
/approved

ebattat (approved):
/approved

PR #1193, opened 2026-04-19
Fix vdbench workload: add ephemeral support and refactor storage type logic

ebattat:
## Summary
- Add `vdbench_pod_ephemeral` and `vdbench_vm_ephemeral` workload support
- Refactor storage type / ephemeral detection logic to handle any workload ending with `_ephemeral` (not just hammerdb/winmssql)
- Add vdbench VM ephemeral template and virtctl integration test

## Test plan
- [ ] Verify `vdbench_pod_ephemeral` workload runs correctly with ephemeral storage
- [ ] Verify `vdbench_vm_ephemeral` workload runs correctly with ephemeral storage
- [ ] Verify existing vdbench_pod / vdbench_vm workloads still work with ODF storage
- [ ] Verify hammerdb ephemeral workloads still work as expected

---
*This PR was created with the assistance of Cursor AI.*

Made with [Cursor](https://cursor.com)

### DIFF
--- a/benchmark_runner/common/template_operations/templates/hammerdb/hammerdb_data_template.yaml
+++ b/benchmark_runner/common/template_operations/templates/hammerdb/hammerdb_data_template.yaml
@@ -29,13 +29,13 @@ template_data:
       db_warehouses: 2
       storage: 10Gi
       database_limits_cpu: 4
-      database_limits_memory: 16Gi
+      database_limits_memory: 12Gi
       database_requests_cpu: 10m
-      database_requests_memory: 16Gi
+      database_requests_memory: 4Gi
       limits_cpu: 4
-      limits_memory: 16Gi
+      limits_memory: 12Gi
       requests_cpu: 10m
-      requests_memory: 16Gi
+      requests_memory: 4Gi
   extra:
     mariadb:
       db_image: mariadb10.5-centos9-container-disk:latest
@@ -69,6 +69,6 @@ template_data:
           vm_requests_memory: 32Gi
         default:
           sockets: 1
-          cores: 4
-          vm_limits_memory: 16Gi
-          vm_requests_memory: 16Gi
+          cores: 2
+          vm_limits_memory: 12Gi
+          vm_requests_memory: 4Gi
--- a/benchmark_runner/common/template_operations/templates/vdbench/internal_data/vdbench_pod_template.yaml
+++ b/benchmark_runner/common/template_operations/templates/vdbench/internal_data/vdbench_pod_template.yaml
@@ -121,6 +121,10 @@ spec:
         - name: vdbench-pod-pvc-claim
         {%- endif %}
           mountPath: "/workload"
+      {%- else %}
+      volumeMounts:
+        - name: vdbench-ephemeral-vol
+          mountPath: "/workload"
       {%- endif %}
       env:
         - name: BLOCK_SIZES
@@ -197,4 +201,8 @@ spec:
       persistentVolumeClaim:
         claimName: vdbench-pod-pvc-claim
     {%- endif %}
+  {%- else %}
+  volumes:
+    - name: vdbench-ephemeral-vol
+      emptyDir: {}
   {%- endif %}
--- a/benchmark_runner/common/template_operations/templates/vdbench/internal_data/vdbench_pod_template.yaml
+++ b/benchmark_runner/common/template_operations/templates/vdbench/internal_data/vdbench_pod_template.yaml
@@ -90,9 +90,6 @@ spec:
     kubernetes.io/hostname: {{ pin_node1 }}
     {%- endif %}
   {%- endif %}
-  {%- if kind == 'kata' %}
-  runtimeClassName: kata
-  {%- endif %}
   containers:
     - name: vdbench-pod
       image: {{ vdbench_image }}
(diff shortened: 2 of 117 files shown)


benchmark_runner/common/template_operations/templates/vdbench/internal_data/vdbench_pod_template.yaml
```diff
@@ -197,4 +201,8 @@ spec:
       persistentVolumeClaim:
         claimName: vdbench-pod-pvc-claim
     {%- endif %}
+  {%- else %}
+  volumes:
+    - name: vdbench-ephemeral-vol
+      emptyDir: {}
```
arpsharm, line 204: When `odf_pvc=False`, the emptyDir volume is added here, but the DataVolume spec section (line 31) renders based on `cluster != "kubernetes"`, not `odf_pvc`. This means for OpenShift clusters with `ODF_PVC=False`, you get both a DataVolume AND an emptyDir — conflicting storage config.

benchmark_runner/workloads/workloads_operations.py
```diff
@@ -314,42 +314,42 @@ def _create_pod_run_artifacts(self, pod_name: str, log_type: str):
...
-                    num = float(value)
-                    line_dict[key] = round(num, 3)
-                elif value == 'n/a':
-                    line_dict[key] = 0.0
-            line_dict['vm_name'] = vm_name
-            workload = self._get_workload_file_name(workload=self._get_run_artifacts_hierarchy(workload_name=workload_name, is_file=True))
-            line_dict['run_artifacts_url'] = os.path.join(self._run_artifacts_url, f'{workload}.tar.gz')
-            result_list.append(dict(line_dict))
-        return result_list
+    # def _create_vm_run_artifacts(self, vm_name: str, start_stamp: str, end_stamp: str, log_type: str):
```
arpsharm: `_create_vm_run_artifacts` is commented out rather than deleted. This is a shared base class, please either delete it entirely or keep it functional.

benchmark_runner/common/template_operations/templates/hammerdb/hammerdb_data_template.yaml
```diff
@@ -29,13 +29,13 @@ template_data:
       db_warehouses: 2
       storage: 10Gi
       database_limits_cpu: 4
-      database_limits_memory: 16Gi
+      database_limits_memory: 12Gi
       database_requests_cpu: 10m
-      database_requests_memory: 16Gi
+      database_requests_memory: 4Gi
```
arpsharm, line 34: `requests_memory` reduced from 16Gi to 4Gi (4x reduction in guaranteed memory). Has this been tested? With a 12Gi/4Gi limit-to-request ratio of 3:1, database pods could get evicted under memory pressure.

benchmark_runner/workloads/vdbench_vm.py
```diff
@@ -26,6 +30,56 @@ def __init__(self):
...
+            candidates = sorted(glob.glob(os.path.join(self._run_artifacts_path, f'{self.__name}_*.yaml')))
+            if candidates:
+                yaml_path = candidates[0]
+        with open(yaml_path, 'r') as f:
+            for line in f:
+                match = re.search(r'export IO_OPERATION=(.+)', line.strip())
+                if match:
+                    io_operation = match.group(1).strip()
+                    return len(io_operation.split(','))
+        return 0
```
arpsharm: `_get_expected_file_count` returns 0 if no `IO_OPERATION` line matches the regex. Then `wait_for_vm_completed_by_file_count` checks `actual_count >= 0`, which would immediately return True (false completion). Should this raise an error or return a minimum of 1 instead?

benchmark_runner/common/virtctl/virtctl.py
```diff
@@ -103,6 +103,74 @@ def _wait_for_file_created(self, vm_name: str, file_path: str, namespace: str =
...
+        )
+        if not result:
+            logger.warning(f'Failed to list {file_type} files on {vm_name}:{remote_dir}')
+            return []
+        local_files = []
+        for filename in result.strip().splitlines():
+            filename = filename.strip()
+            if not filename:
+                continue
+            remote_path = f'{remote_dir}{filename}'
```
arpsharm: No path separator between `remote_dir` and `filename`. Currently works because all callers pass `/workload/` (with trailing slash), but would break if someone passes `/workload`. Consider using `os.path.join(remote_dir, filename)`.

arpsharm (changes requested):
A few items to address below.

arpsharm (approved):

benchmark_runner/workloads/workloads_operations.py
ebattat: It is related to old parser from logs so I have deleted it.

benchmark_runner/common/template_operations/templates/hammerdb/hammerdb_data_template.yaml
ebattat: Yes, I have tested it for both Pod and VM for functional test, reduce it that it will work against our test cluster.

benchmark_runner/workloads/vdbench_vm.py
ebattat: Thx, goot catch raise error MissingIOOperation

benchmark_runner/common/virtctl/virtctl.py
ebattat: Good catch, fixed !!!

benchmark_runner/common/template_operations/templates/vdbench/internal_data/vdbench_pod_template.yaml
ebattat: You right, fixed

arpsharm (approved):

arpsharm (approved):

PR #1209, opened 2026-04-27
Fix bootstorm VM scale: add process join timeout to prevent workflow hangs

ebattat:
## Summary
- `subprocess.getoutput()` used by `virtctl stop` (and other commands) has no timeout, so `p.join()` blocks forever when the command hangs at scale (100+ VMs)
- Add `OC.SHORT_TIMEOUT` (600s) to `p.join()` and terminate hung child processes
- Prevents the workflow from getting stuck on the stop step when `virtctl stop` hangs due to API server overload

## Test plan
- [ ] Run Windows VM bootstorm at scale (100+ VMs) and verify workflow no longer gets stuck on stop step
- [ ] Verify terminated processes are logged with `timed out in <step>, terminated`
- [ ] Verify normal (non-hanging) runs are unaffected by the timeout

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * VM workload benchmarks now enforce execution timeouts to prevent indefinite hangs and automatically terminate runaway processes, with a grace period before forceful kill.
  * Error messages enhanced to include the affected process and the active workload step, improving diagnosability and monitoring of timeout events.

### DIFF
--- a/benchmark_runner/workloads/bootstorm_vm.py
+++ b/benchmark_runner/workloads/bootstorm_vm.py
@@ -6,7 +6,7 @@ from multiprocessing import Process, Manager
 from benchmark_runner.common.logger.logger_time_stamp import logger_time_stamp, logger
 from benchmark_runner.common.elasticsearch.elasticsearch_exceptions import ElasticSearchDataNotUploaded
 from benchmark_runner.workloads.workloads_operations import WorkloadsOperations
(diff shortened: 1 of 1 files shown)


ebattat:
I will remove win10 who cause this issue, it EOL.

PR #1213, opened 2026-04-30
Replace stressng/uperf pod images with Fedora 43

arpsharm:
## Summary
- Replace old UBI 8 pod images (`stressng:latest`, `uperf:latest`) with single pre-built `quay.io/benchmark-runner/fedora:43`
- New Dockerfile at `dockerfiles/fedora43-pod/` with pinned versions: stress-ng-0.20.01, uperf-1.0.8, python3-pyyaml-6.0.2
- Add `FEDORA` env variable for pod image reference (single place to update)
- Add `stressng_version` and `uperf_version` tracking to ElasticSearch metadata

### DIFF
--- a/benchmark_runner/common/template_operations/templates/stressng/internal_data/stressng_pod_template.yaml
+++ b/benchmark_runner/common/template_operations/templates/stressng/internal_data/stressng_pod_template.yaml
@@ -54,8 +54,11 @@ spec:
       {%- if kind == 'kata' %}
       runtimeClassName: kata
       {%- endif %}
+      serviceAccountName: {{ namespace }}
       containers:
       - name: stressng
+        securityContext:
+          runAsUser: 0
         {%- if resources == 'true' or resources == true %}
         resources:
           requests:
@@ -65,7 +68,7 @@ spec:
             cpu: {{ limits_cpu }}
             memory: {{ limits_memory }}
         {%- endif %}
-        image: {{ image | default('quay.io/benchmark-runner/stressng:latest') }}
+        image: {{ image | default('quay.io/fedora/fedora:43') }}
         imagePullPolicy: Always
         env:
           - name: uuid
@@ -106,6 +109,7 @@ spec:
         args:
           - |
             set -e
+            dnf install -y stress-ng python3-pyyaml > /dev/null 2>&1
             cd /tmp
             stress-ng --job /workload/jobfile --log-file /tmp/stressng.log -Y /tmp/stressng.yml || exit 1
             # Parse results into JSON (visible in pod logs, uploaded by benchmark-runner)
--- a/benchmark_runner/common/template_operations/templates/stressng/internal_data/stressng_pod_template.yaml
+++ b/benchmark_runner/common/template_operations/templates/stressng/internal_data/stressng_pod_template.yaml
@@ -54,11 +54,8 @@ spec:
       {%- if kind == 'kata' %}
       runtimeClassName: kata
       {%- endif %}
-      serviceAccountName: {{ namespace }}
       containers:
       - name: stressng
-        securityContext:
-          runAsUser: 0
         {%- if resources == 'true' or resources == true %}
         resources:
           requests:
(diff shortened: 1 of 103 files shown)


tests/integration/benchmark_runner/templates/uperf_pod_client_template.yaml
```diff
@@ -14,9 +14,12 @@ spec:
         app: uperf-bench-client
         type: uperf-bench-client-test123
     spec:
+      serviceAccountName: benchmark-runner
       containers:
       - name: benchmark
-        image: quay.io/benchmark-runner/uperf:latest
+        image: quay.io/fedora/fedora:43
```
ebattat: Using the upstream Fedora image directly is risky for benchmarks because the tag can change over time, which means the environment is not fixed and our performance results may not be reproducible or comparable between runs.
The best practice in to upload it to:
quay.io/benchmark-runner/fedora:43

tests/unittest/benchmark_runner/common/template_operations/golden_files/chaos_ci_stressng_pod_ODF_PVC_False/stressng_pod.yaml
```diff
@@ -81,6 +84,7 @@ spec:
         args:
           - |
             set -e
+            dnf install -y stress-ng python3-pyyaml > /dev/null || exit 1
```
ebattat: can we install specific version of stress-ng because installing latest can affect the results ?
if yes let upload the version number to elastic stress-ng_version

tests/unittest/benchmark_runner/common/template_operations/golden_files/chaos_ci_uperf_kata_ODF_PVC_False/uperf_pod_server.yaml
```diff
@@ -136,12 +136,15 @@ spec:
...
       containers:
       - name: benchmark
-        image: quay.io/benchmark-runner/uperf:latest
+        image: quay.io/fedora/fedora:43
+        securityContext:
+          runAsUser: 0
         imagePullPolicy: Always
         command: ["/bin/sh", "-c"]
-        args: ["uperf -s -v -P 30000"]
+        args: ["(dnf install -y uperf > /dev/null || exit 1) && uperf -s -v -P 30000"]
```
ebattat: can we install specific version of uperf because installing latest can affect the results ?
if yes let upload the version number to elastic uperf_version

benchmark_runner/common/template_operations/templates/stressng/internal_data/stressng_pod_template.yaml
```diff
@@ -65,7 +65,7 @@ spec:
             cpu: {{ limits_cpu }}
             memory: {{ limits_memory }}
         {%- endif %}
-        image: {{ image | default('quay.io/benchmark-runner/stressng:latest') }}
+        image: {{ image | default('quay.io/benchmark-runner/fedora:43') }}
```
ebattat: @arpsharm,
can you create env variable:
        self._environment_variables_dict['fedora'] = EnvironmentVariables.get_env(
            'FEDORA', 'quay.io/benchmark-runner/fedora:43')
so next time we will update only 1 place and not several ?

dockerfiles/fedora43-pod/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM quay.io/fedora/fedora:43
+RUN dnf install -y stress-ng python3-pyyaml uperf && dnf clean all
```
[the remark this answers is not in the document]
ebattat: You always install the latest stressng and uperf version:
dnf install -y stress-ng
while you mention hard coded values: 
            'stressng': '0.20.01',
            'uperf': '1.0.8'

I think you should install like this:
dnf install -y stress-ng-0.20.01 uperf-1.0.8
btw, what we install inside the fedora vm image ?

btw, in benchmark-operator uperf/stressng used the latest version ?

arpsharm:
Hi @ebattat so for VMs we use the latest version for stressng and uperf in the image, currently which is stress-ng 0.20.01 and uperf 1.0.8. As for benchmark-operator, it also installed the latest version for stressng and uperf as well but since the image hasn't been rebuilt for a couple of years I am guessing the version on it would be old to what we have today.

ebattat (approved):
/approved

PR #1216, opened 2026-05-05
Fix Python version detection to skip unreleased versions

arpsharm:
## Summary
- Fix get_latest_python_patch.py to validate the source tarball exists via HTTP HEAD before accepting a version
- Prevents CI failures when python.org creates a directory for an RC but the final tarball isn't released yet
- Currently broken: script returns 3.14.5 (RC directory exists) but Python-3.14.5.tgz doesn't exist, causing 404 in Dockerfile build
- With fix: correctly returns 3.14.4 (latest version with actual tarball)

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable selection of Python patch releases by confirming source tarballs are reachable before choosing a patch.
  * Clearer, consistent error reporting when parsing fails or when no suitable patches are found.

* **Refactor**
  * Improved discovery logic to gather and validate multiple candidate patches for more robust selection.

### DIFF
--- a/hack/get_latest_python_patch.py
+++ b/hack/get_latest_python_patch.py
@@ -2,65 +2,77 @@
 """
 Find the latest point release for a given Python minor version (e.g. 3.14)
 by parsing https://www.python.org/ftp/python/
+Validates that the source tarball actually exists (not just an RC directory).
 Used at Docker build time to install the latest bugfix release.
 """
 from html.parser import HTMLParser
+import logging
 import sys
 import urllib.request
 import urllib.error
+
+logger = logging.getLogger(__name__)
 
 URL = 'https://www.python.org/ftp/python/'
 
 
 def find_best_python_version(release):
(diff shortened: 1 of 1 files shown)


ebattat (approved):
/approved

hack/get_latest_python_patch.py
```diff
@@ -46,18 +43,31 @@ def get_best_version(self):
...
     try:
         parser.feed(data)
-        if parser.get_best_version() is not None:
-            print(parser.get_best_version())
-        else:
-            print(f'Cannot find python version for {release}', file=sys.stderr)
-            sys.exit(1)
     except Exception as e:
-        print(f'Cannot find python version for {release}: {e}', file=sys.stderr)
+        print(f'Cannot parse python versions for {release}: {e}', file=sys.stderr)
```
ebattat: @arpsharm pls change print to logger

PR #1217, opened 2026-05-05
Remove stressng/uperf version pinning

arpsharm:
## Summary
- Remove version pinning from Dockerfile, install latest stress-ng/uperf/python3-pyyaml
- Matches benchmark-operator behavior of always using latest versions

## Summary by CodeRabbit

* **Chores**
  * Updated Docker image configuration to install latest available versions of development tools instead of specific pinned versions.


ebattat (approved):
/approved

PR #1221, opened 2026-05-11
Support HammerDB flat run using the HAMMERDB_CONFIG environment variable

ebattat:
## Summary
- Add optional `HAMMERDB_CONFIG` env var to override HammerDB template defaults at runtime
- Supports: `db_min_workers`, `db_num_workers`, `db_warehouses`, `runtime`, `rampup`, `iterations`, `transactions`
- For flat run (no ramp-up): set `db_min_workers` = `db_num_workers`
- Fix hardcoded `rampup`/`runtime` in winmssql `05_hammerdb-auto-runs` script (was always 1)
- Add SSH key verification with diagnostic message for cloudbase-init failures
- Add `WINDOWS_URL` accessibility check at startup

## Usage
```bash
# test_ci defaults (no override needed):
# db_min_workers=1, db_num_workers=2, db_warehouses=2, runtime=1, rampup=1, iterations=2

# Flat 32-thread, 60 min run:
HAMMERDB_CONFIG="{'db_min_workers': 32, 'db_num_workers': 32, 'runtime': 60, 'iterations': 1}"

# Custom warehouses + duration:
HAMMERDB_CONFIG="{'db_warehouses': 192, 'runtime': 120}"
```

## Test plan
- [x] Verify default behavior unchanged (no HAMMERDB_CONFIG set)
- [x] Verify flat run with db_min_workers = db_num_workers
- [x] Verify runtime override renders correctly in generated scripts
- [x] Golden file tests pass

Assisted-by: Claude Code

## Summary by CodeRabbit

* **New Features**
  * Environment-variable override for HammerDB template data
  * New template parameters to configure minimum workers and iteration count

* **Improvements**
  * Workload generation respects configurable min-worker and iteration settings
  * Windows VM initialization: URL reachability check, clearer SSH/setup guidance, and non-fatal SSH key presence warning

[Review Change Stack](https://app.coderabbit.ai/change-stack/redhat-performance/benchmark-runner/pull/1221?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/benchmark_runner/common/template_operations/template_operations.py
+++ b/benchmark_runner/common/template_operations/template_operations.py
@@ -150,6 +150,11 @@ class TemplateOperations:
 
         render_data = self.__build_template_data(template_render_data, workload_data)
 
+        hammerdb_config = self.__environment_variables_dict.get('hammerdb_config', {})
+        if hammerdb_config:
+            logger.info(f'HAMMERDB_CONFIG override: {hammerdb_config}')
+            render_data.update(hammerdb_config)
+
         out_files = []
         standard_template_path = os.path.join(workload_dir_path, 'internal_data', self.__standard_template_file)
         if os.path.isfile(standard_template_path):
--- a/benchmark_runner/common/template_operations/template_operations.py
+++ b/benchmark_runner/common/template_operations/template_operations.py
@@ -153,6 +153,14 @@ class TemplateOperations:
 
         render_data = self.__build_template_data(template_render_data, workload_data)
 
+        hammerdb_config = self.__environment_variables_dict.get('hammerdb_config', {})
+        if hammerdb_config and 'hammerdb' in self.__workload_name:
+            unknown_keys = set(hammerdb_config.keys()) - set(render_data.keys())
+            if unknown_keys:
+                logger.warning(f'HAMMERDB_CONFIG unknown keys (will be ignored): {unknown_keys}')
+            logger.info(f'HAMMERDB_CONFIG override: {hammerdb_config}')
+            render_data.update(hammerdb_config)
+
         out_files = []
         standard_template_path = os.path.join(workload_dir_path, 'internal_data', self.__standard_template_file)
         if os.path.isfile(standard_template_path):
--- a/benchmark_runner/common/template_operations/templates/hammerdb/hammerdb_data_template.yaml
+++ b/benchmark_runner/common/template_operations/templates/hammerdb/hammerdb_data_template.yaml
@@ -9,6 +9,8 @@ template_data:
     pin_node2: "{{ pin_node2 }}"
     odf_pvc: {{ odf_pvc }}
     transactions: 100000
+    db_min_workers: 1
+    iterations: 2
     rampup: 1
     runtime: 1
   run_type:
--- a/benchmark_runner/common/template_operations/templates/hammerdb/internal_data/hammerdb_mariadb_configmap.yaml
+++ b/benchmark_runner/common/template_operations/templates/hammerdb/internal_data/hammerdb_mariadb_configmap.yaml
@@ -94,7 +94,7 @@ data:
     set workers {{ db_num_workers }}
     set transactions {{ transactions }}
     set samples 1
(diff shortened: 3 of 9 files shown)


ebattat (commented):
[Claude Code] Reviewed all 5 CodeRabbit findings. Here's what was addressed:

## Fixed

**#1 — VU doubling can skip peak (winmssql PowerShell template)**
Added a guard to ensure `db_min_workers <= db_num_workers` (fail-fast with clear error) and appended `$maxVU` to `$vuCounts` if the doubling loop didn't land on it. This prevents the peak VU count from being silently skipped when doubling doesn't align exactly.

The same doubling pattern exists in the 3 hammerdb Tcl configmaps (mariadb, mssql, postgres), but those are not affected — `db_min_workers` and `db_num_workers` are always powers of 2 (1→2→4→8→16→32), so the loop always lands on the peak.

## Skipped (with reasoning)

**#2 — `ast.literal_eval` crash on malformed `HAMMERDB_CONFIG`**: Crashing on invalid input is the correct behavior. A silent fallback to `{}` would hide the misconfiguration and run benchmarks with wrong defaults, which is worse than failing loudly.

**#3 — URL scheme validation for `WINDOWS_URL`**: The URL comes from an operator-controlled environment variable, not untrusted user input. Scheme validation adds complexity without meaningful security benefit in this context.

**#4 — Scope `hammerdb_config` to hammerdb workloads only**: Jinja2 ignores unused variables in `render_data`. Extra keys from `hammerdb_config` are harmless when applied to non-hammerdb workloads.

**#5 — `iterations < 1` guard**: The value comes from template defaults (2) or explicit `HAMMERDB_CONFIG` override. Setting `iterations=0` is user error, not a code defect.

arpsharm (commented):
[render_data.update(hammerdb_config)](https://github.com/redhat-performance/benchmark-runner/blob/fabe3b510e49df5b84dca5fd88e1aabe67fa10dd/benchmark_runner/common/template_operations/template_operations.py#L159) accepts any keys without validation. A typo like db_nmu_workers instead of db_num_workers silently falls through to defaults. Worth adding a warning for unknown keys?

ebattat:
@arpsharm, I don't think we should do key validation, the user should verify that they entered the correct values.

ebattat (commented):
[Claude Code] Addressed CodeRabbit review findings:

## Fixed

**Hardcoded username in SSH diagnostic message (`winmssql_vm.py:87`)**
Replaced hardcoded `'Administrator'` with `self.__username` so the diagnostic message uses the configured VM user.

## Skipped (with reasoning)

**`save_error_logs` in child processes (`winmssql_vm.py:84-95`)**: This is the existing pattern used across all workloads (vdbench, bootstorm, hammerdb). Changing it only for winmssql would be inconsistent. If this needs to change, it should be a separate refactoring PR across all workloads.

**URL scheme validation (`winmssql_vm.py:24-27`)**: The URL comes from an operator-controlled environment variable, not untrusted user input. Adding scheme validation adds complexity without meaningful security benefit in this context.

arpsharm (commented):
[render_data.update(hammerdb_config)](https://github.com/redhat-performance/benchmark-runner/blob/5f9a9b377e9cce5950c04e16973440b96b8d2523/benchmark_runner/common/template_operations/template_operations.py#L159) applies to all workloads. If HAMMERDB_CONFIG is set while running a non-hammerdb workload, keys leak into that template context. Should guard with an if-hammerdb check.

ebattat:
[Claude Code] Fixed in c30a20b5 — added a warning for unknown keys in HAMMERDB_CONFIG so typos like `db_nmu_workers` are surfaced in the logs instead of silently falling through to defaults.

ebattat:
[Claude Code] Fixed in c30a20b5 — added `and 'hammerdb' in self.__workload_name` guard so HAMMERDB_CONFIG only applies when running a hammerdb workload, preventing keys from leaking into other template contexts.

arpsharm (approved):
/lgtm — both previous review comments addressed (unknown key validation + hammerdb guard). Minor nit: unused 'from multiprocessing import Process' import in winmssql_vm.py.

arpsharm (approved):

PR #1223, opened 2026-05-12
Use workload timeout for bootstorm VM access wait

ebattat:
## Summary
- Pass `self._timeout` (from `TIMEOUT` env var, default 3600s) to `wait_for_vm_access` instead of hardcoded `SHORT_TIMEOUT` (600s)
- Windows Server 2025 VMs at scale (37+ per worker) need more than 10 minutes to boot when competing for I/O
- Only 59 out of 111 VMs were accessible within the 600s timeout; increasing to 3600s allows slower VMs to complete boot

## Test plan
- [ ] Run Windows Server 2025 bootstorm with 37 VMs per worker (3 workers)
- [ ] Verify all 111 VMs become accessible within the extended timeout
- [ ] Verify Windows Server 2022 bootstorm still works (regression check)

Assisted-by: Claude Code

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved VM startup timeout handling in the bootstorm workload: explicit timeout is now honored, timeout failures are caught and logged as warnings, and VM access attempts are recorded as failed when timed out—leading to more consistent, reliable behavior during VM startup operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/redhat-performance/benchmark-runner/pull/1223)

### DIFF
--- a/benchmark_runner/workloads/bootstorm_vm.py
+++ b/benchmark_runner/workloads/bootstorm_vm.py
@@ -7,6 +7,7 @@ from benchmark_runner.common.logger.logger_time_stamp import logger_time_stamp, 
 from benchmark_runner.common.elasticsearch.elasticsearch_exceptions import ElasticSearchDataNotUploaded
 from benchmark_runner.workloads.workloads_operations import WorkloadsOperations
 from benchmark_runner.common.oc.oc import VMStatus
+from benchmark_runner.common.oc.oc_exceptions import VMStateTimeout
 
 
 class BootstormVM(WorkloadsOperations):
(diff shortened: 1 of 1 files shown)


arpsharm (commented):
When [VMStateTimeout is caught](https://github.com/redhat-performance/benchmark-runner/blob/aca5fa4b09edd798536a45a62ac6afadc86d7583/benchmark_runner/workloads/bootstorm_vm.py#L80), vm_node is still truthy so [access_vm=int(bool(vm_node))](https://github.com/redhat-performance/benchmark-runner/blob/aca5fa4b09edd798536a45a62ac6afadc86d7583/benchmark_runner/workloads/bootstorm_vm.py#L111) records access_vm=1 (success) even though access failed. Should set vm_node=False or add a separate flag before return.

PR #1255, opened 2026-07-27
Per-parameter Orion regression detection for all workloads

arpsharm:
- Split hammerdb into per-VU tests (1,2,4,8,16,32) × 13 variants = 78 tests
- Split fio by block_size (4k,8k,32k,1M) × io_operation (read,write,randread,randwrite) × kind = 32 tests
- Split uperf by num_threads (1,8) × message_size (64,1024,8192) × test_type (stream,rr) × kind = 24 tests
- Split vdbench by Run profile (14 profiles) × kind = 28 tests
- Split sysbench by test_type (cpu,memory) × kind = 4 tests
- Fix winmssql metric (bootstorm_time → tpm) and Jenkinsfile ES index (windows-results → hammerdb-results)
- Fix winfio bandwidth field (bandwidth_MiBsec → total_bw_kbs)
- Drop vm_os_version filters from winmssql/winfio
- Add winmssql/winfio to run-orion-all.sh

## Summary by CodeRabbit

- **New Features**
  - Expanded performance test coverage across uperf, fio, HammerDB, sysbench, and vdbench for both VM and pod environments.
  - Added additional uperf test matrix variants (stream and rr) across thread counts and message sizes.
  - Split sysbench into dedicated CPU and memory test cases.

- **Updates**
  - Enhanced fio, HammerDB, and vdbench configs with more explicitly defined scenarios and standardized thresholds/metrics.
  - Updated WinFIO to report total bandwidth and WinMSSQL to use TPM instead of bootstorm timing. 
  - Added additional workloads to Orion regression analysis and reporting outputs.

### DIFF
--- a/orion-configs/perfci-hammerdb.yaml
+++ b/orion-configs/perfci-hammerdb.yaml
@@ -1,10 +1,11 @@
 tests:
-  - name: hammerdb-vm-mssql-odf
+  - name: hammerdb-vm-mssql-odf-1vu
     metadata:
       kind: vm
       run_status: complete
       db_type: mssql
       storage_type: odf
+      current_worker: 1
     metrics:
       - name: tpm
         metric_of_interest: tpm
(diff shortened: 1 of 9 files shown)


orion-configs/perfci-hammerdb.yaml
ebattat, line 1: @arpsharm, there is a lot of metric to review
maybe we should keep it as it was.
If there is a degradation we can drill down inside grafana
So, let keep using aggregate TPM per database

arpsharm: I think its good because now it can track every variant, their subtest types with different blocks and sizes. It's big but not much complex to understand because the key metrics are the same for those workload variants and subtypes. However, upto you if you think good to merge it

ebattat (approved):
/approve

PR #1263, opened 2026-08-03
Fix inverted Orion direction values across all perfci configs

arpsharm:
All perfci Orion configs had direction values inverted so throughput/IOPS/tpm were flagging increases as regressions and latency/time were flagging decreases as regressions.

- `direction: 1` → increase is regression (latency, boot time, CPU)
- `direction: -1` → decrease is regression (throughput, IOPS, tpm, bandwidth)

Also adds missing vdbench variant `64kb_write_16threads` (vm + pod).

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected performance threshold evaluation across boot, database, storage, CPU, memory, and networking benchmarks.
  - Throughput, IOPS, bandwidth, and operations metrics now use the appropriate comparison direction, while latency metrics are evaluated correctly.
  - Updated performance checks across Linux and Windows environments, including Windows 11, Server 2022, and Server 2025.

- **New Features**
  - Added 64 KB write testing with 16 threads to Vdbench performance coverage for both virtual machines and pods.

### DIFF
--- a/orion-configs/perfci-bootstorm.yaml
+++ b/orion-configs/perfci-bootstorm.yaml
@@ -9,10 +9,10 @@ tests:
         agg:
           agg_type: avg
         threshold: 10
-        direction: -1
+        direction: 1
       - name: total-run-time
         metric_of_interest: total_run_time
         agg:
           agg_type: avg
         threshold: 10
-        direction: -1
+        direction: 1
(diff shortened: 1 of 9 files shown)


ebattat (approved):
/approve

PR #1270, opened 2026-08-10
Add CLEAR_CACHE flag to support scale env.

ebattat:
## Summary

- Add \`CLEAR_CACHE\` environment variable (default: \`True\`) to support scale environments
- Decouple \`clear_nodes_cache()\` from \`DELETE_ALL\` in \`workloads_operations.py\` and \`clusterbuster_workloads.py\` so each can be controlled independently

## Motivation

Previously \`clear_nodes_cache()\` was always called together with \`delete_all()\` — it was not possible to skip cache clearing without also skipping resource deletion. This adds flexibility for re-running workloads on an already-prepared cluster to save time.

This improvement is particularly useful **for support scale env.** — in large-scale runs where many VMs/pods are deployed across multiple nodes, clearing the cache on every node adds significant overhead. Setting \`CLEAR_CACHE=False\` allows scale workloads to skip this step and reduce total run time.

## Usage

\`\`\`bash
# Skip cache clearing (default is True)
CLEAR_CACHE=False python -m benchmark_runner.main.main
\`\`\`

## Test plan
- [x] All 26 unit tests pass
- [x] Applied to both \`workloads_operations.py\` and \`clusterbuster_workloads.py\`

🤖 Assisted-by: Claude Code

### DIFF
--- a/benchmark_runner/clusterbuster/clusterbuster_workloads.py
+++ b/benchmark_runner/clusterbuster/clusterbuster_workloads.py
@@ -66,7 +66,8 @@ class ClusterBusterWorkloads(WorkloadsOperations):
         :return:
         """
         self.delete_all()
-        self.clear_nodes_cache()
+        if self._clear_cache:
+            self.clear_nodes_cache()
         if self._enable_prometheus_snapshot:
             self.start_prometheus()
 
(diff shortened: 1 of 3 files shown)


ebattat:
[Claude Code] Fixed — applied `CLEAR_CACHE` guard to `clusterbuster_workloads.py` as well, so `clear_nodes_cache()` is now conditional on `self._clear_cache` in both the base workload initializer and the ClusterBuster workload.

PR #1277, opened 2026-08-22
Fix QUAY_AUTH encoding: pipe credentials via stdin to avoid quoting issues

ebattat:
## Summary
- Fix `NameError: name 'QUAY_USERNAME' is not defined` in ODF Registered Jenkins stage
- The previous approach using `os.environ[\"QUAY_USERNAME\"]` inside a single-quoted `sh '''` block caused quoting conflicts
- Replace with `printf '%s:%s' "$QUAY_USERNAME" "$QUAY_PASSWORD" | python3 -c "import base64, sys; ..."` — credentials piped via stdin, never exposed as process arguments

## Test
Tested on cluster — `QUAY_AUTH` encodes correctly and `Login Succeeded!`

🤖 Assisted-by: Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved registry authentication handling in the performance CI deployment process.
  * Enhanced storage health monitoring by allowing additional time for recovery before restarting a manager component when health remains degraded.
  * Added clearer warnings when recovery actions are triggered during storage health checks.

### DIFF
--- a/jenkins/PerfCI/02_PerfCI_Operators_Deployment/Jenkinsfile
+++ b/jenkins/PerfCI/02_PerfCI_Operators_Deployment/Jenkinsfile
@@ -90,7 +90,7 @@ END
                             trap "rm -f global-pull-secret.json global-pull-secret.json.tmp" EXIT
                             install -m 600 /dev/null global-pull-secret.json
                             oc get secret pull-secret -n openshift-config -o json | jq -r '.data.".dockerconfigjson"' | base64 -d > global-pull-secret.json
-                            QUAY_AUTH=$(QUAY_USERNAME="$QUAY_USERNAME" QUAY_PASSWORD="$QUAY_PASSWORD" python3 -c "import base64, os; print(base64.b64encode(f'{os.environ[\"QUAY_USERNAME\"]}:{os.environ[\"QUAY_PASSWORD\"]}'.encode()).decode(), end='')")
+                            QUAY_AUTH=$(printf '%s:%s' "$QUAY_USERNAME" "$QUAY_PASSWORD" | python3 -c "import base64, sys; print(base64.b64encode(sys.stdin.buffer.read()).decode(), end='')")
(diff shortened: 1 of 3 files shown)


jenkins/PerfCI/02_PerfCI_Operators_Deployment/Jenkinsfile
```diff
@@ -90,7 +90,7 @@ END
                             trap "rm -f global-pull-secret.json global-pull-secret.json.tmp" EXIT
                             install -m 600 /dev/null global-pull-secret.json
                             oc get secret pull-secret -n openshift-config -o json | jq -r '.data.".dockerconfigjson"' | base64 -d > global-pull-secret.json
-                            QUAY_AUTH=$(QUAY_USERNAME="$QUAY_USERNAME" QUAY_PASSWORD="$QUAY_PASSWORD" python3 -c "import base64, os; print(base64.b64encode(f'{os.environ[\"QUAY_USERNAME\"]}:{os.environ[\"QUAY_PASSWORD\"]}'.encode()).decode(), end='')")
+                            QUAY_AUTH=$(printf '%s:%s' "$QUAY_USERNAME" "$QUAY_PASSWORD" | python3 -c "import base64, sys; print(base64.b64encode(sys.stdin.buffer.read()).decode(), end='')")
```
RobertKrawitz, line 93: This isn't the right way to do it; QUAY_PASSWORD still exists on a command line (to printf).  The right way to do this is:

```
QUAY_AUTH=$(printf '%s:%s' | python3 -c "import base64, sys; print(base64.b64encode(sys.stdin.buffer.read()).decode(), end='')" <<< "${QUAY_USERNAME}:${QUAY_PASSWORD}" )
```

ebattat: All the passowrd are under secret and this jenkins run under VPN, I dont think it is issue

